PR #14, opened 2024-07-25
Update README.md to add Scorecard as a pilot project

Danajoyluck:
Met with Scorecard maintainers and they would like to be part of the  pilot while applying for incubating lifecycle change.


eddie-knight (approved):
LGTM, excited to have Scorecard on board for the pilot

SecurityCRob (approved):

PR #31, opened 2024-10-17
Adding entries for probes that already exist

eddie-knight:
Also added notes for places where we making probes might not be straightforward, and removed the key for probe or security insights in places where that key definitely isn't needed.

Also updated the template to handle one or multiple probes. And incidentally fixed some trailing spaces.


SecurityCRob (approved):
noice

TheFoxAtWork:
@eddie-knight any reason this cannot be merged?

eddie-knight:
Thanks @TheFoxAtWork ✊

PR #33, opened 2024-10-18
fix: clarify definition of a Collaborator

madpah:
MFA could not be enforced on committers that contribute to a PR that originates from a Fork.

@eddie-knight


TheFoxAtWork (approved):
LGTM - generally i think this could be better defined or provide examples to indicate unblocked write access to the repo and administrator privileges of the repo (if that is what we mean).

eddie-knight (approved):

PR #34, opened 2024-10-18
Update baseline.md Criteria Table

SecurityCRob:
massage order of items in criteria table to group like-items together.  I did not yet adjust any IDs and await group feedback prior to making further adjustments.


TheFoxAtWork (approved):
overall this looks fine and it makes sense to reorganize. Perhaps it may be beneficial to alter the numbering scheme to reflect the category?
OSPS-AC-1.1
OSPS-AC-1.2
OSPS-AC-1.3
OSPS-AC-2.1
OSPS-AC-2.2

in doing this method, it allows for sorting by category. the numerical value also indicates the level and component under that level we're referring to. further, we can continue to add criteria for any we miss without having to rework criteria numbering and reorganization. this means in the event new categories need added, we have a schema for them.

PR #35, opened 2024-10-18
added two new criteria

SecurityCRob:
add osps-53 & osps-54 per convo with eknight

### DIFF
--- a/baseline.yaml
+++ b/baseline.yaml
@@ -756,10 +756,11 @@ criteria:
     maturity_level: 3
     category: Quality
     criteria: |
-      All changes to the project's codebase MUST
-      require a passing status check from a
-      Static Application Security Testing (SAST)
-      tool.
+      All proposed changes to the project's
+      codebase must be automatically evaluated 
+      against a documented policy for known
+      vulnerabilities and blocked in the
+      event of violations.
     objective: |
       Identify and address defects and security weaknesses 
       in the project's codebase early in the 
--- a/baseline.yaml
+++ b/baseline.yaml
@@ -738,6 +738,12 @@ criteria:
       time with a cryptographic signature or attestations, such 
       as GPG or PGP signature, Sigstore signatures, SLSA provenance, or SLSA VSAs.   
       
+      Provide clear documentation for users on how they can verify the signature
+      and what a valid signature or attestation indicates (e.g.
+      "it came from our CI/CD system",
+      "one of our project maintainers built the software", "it was built from the
+      indicated source repo with the indicated builder", "it meets SLSA level 3").
+      
       It is recommended that this signature is
       generated as part of the build and release
       pipeline to ensure that it is consistent and
--- a/baseline.yaml
+++ b/baseline.yaml
@@ -726,7 +726,7 @@ criteria:
     maturity_level: 2
     category: Build & Release
     criteria: |
-      All released software assets MUST be signed
+      All released software assets MUST be signed or accounted for in a signed manifest including each asset's cryptographic hashes (such as an SBOM, provenance attestation or similar). The identity of the person or process authoring the software release must be published, discoverable and must be taken into account during the verification process.
       with a cryptographic signature.    
     objective: |
       Provide users with a mechanism to verify the
--- a/baseline.yaml
+++ b/baseline.yaml
@@ -735,8 +735,8 @@ criteria:
       tampering or unauthorized modifications.  
     implementation: |
       Sign all released software assets at build 
-      time with a cryptographic signature, such 
-      as a GPG or PGP signature.   
+      time with a cryptographic signature or attestations, such 
+      as GPG or PGP signature, Sigstore signatures, SLSA provenance, or SLSA VSAs.   
       
       It is recommended that this signature is
       generated as part of the build and release


baseline.yaml
```diff
@@ -720,6 +720,55 @@ criteria:
       in a separate repository and fetched during
       a specific well-documented pipeline step.
     control_mappings: # TODO
+    security_insights_value: # TODO
+    scorecard_probe: # TODO
+  - id: OSPS-53
```
mlieberman85, line 725: I disagree with this one as worded fairly strongly. 

I would reword it "cryptographically attested" -- We've seen multiple issues with just signing a piece of software.

baseline.yaml
```diff
@@ -720,6 +720,55 @@ criteria:
...
+    maturity_level: 2
+    category: Build & Release
+    criteria: |
+      All released software assets MUST be signed
```
mlieberman85, line 729: This is not nearly rigorous enough. Anyone can sign anything. The goal here is to cryptographically tie some element of the software (e.g. it's build) to the approved maintainers of that software. As written it's way too weak.

baseline.yaml
```diff
@@ -720,6 +720,55 @@ criteria:
...
+    control_mappings: # TODO
+    security_insights_value: # TODO
+    scorecard_probe:
+      - releasesAreSigned 
+  - id: OSPS-54
+    maturity_level: 3
+    category: Quality
+    criteria: |
+      All changes to the project's codebase MUST
+      require a passing status check from a
```
mlieberman85, line 754: This is too vague. Who decides what the passing status check is? 

This is why I like SSDF's PW7 and PW8

> Review and/or Analyze Human-Readable
Code to Identify Vulnerabilities and Verify
Compliance with Security Requirements
(PW.7): Help identify vulnerabilities so that
they can be corrected before the software is
released to prevent exploitation. Using
automated methods lowers the effort and
resources needed to detect vulnerabilities.
Human-readable code includes source code,
scripts, and any other form of code that an
organization deems human-readable.

> Test Executable Code to Identify
Vulnerabilities and Verify Compliance with
Security Requirements (PW.8): Help identify
vulnerabilities so that they can be corrected
before the software is released in order to
prevent exploitation. Using automated
methods lowers the effort and resources
needed to detect vulnerabilities and improves
traceability and repeatability. Executable code
includes binaries, directly executed bytecode
and source code, and any other form of code
that an organization deems executable.

SSDF instead of explicitly calling out SAST states it a bit more broadly and also highlights that it's up to the organization or system's security needs to determine what to do when an issue arises.

Other docs like NIST 800-53 include similar but a bit more specific wording, e.g.

> ) DEVELOPER TESTING AND EVALUATION | STATIC CODE ANALYSIS
Require the developer of the system, system component, or system service to employ
static code analysis tools to identify common flaws and document the results of the
analysis.
Discussion: Static code analysis provides a technology and methodology for security reviews
and includes checking for weaknesses in the code as well as for the incorporation of libraries
or other included code with known vulnerabilities or that are out-of-date and not supported.
Static code analysis can be used to identify vulnerabilities and enforce secure coding
practices. It is most effective when used early in the development process, when each code
change can automatically be scanned for potential weaknesses. Static code analysis can
provide clear remediation guidance and identify defects for developers to fix. Evidence of
the correct implementation of static analysis can include aggregate defect density for critical
defect types, evidence that defects were inspected by developers or security professionals,
and evidence that defects were remediated. A high density of ignored findings, commonly
referred to as false positives, indicates a potential problem with the analysis process or the
analysis tool. In such cases, organizations weigh the validity of the evidence against evidence
from other sources.

I would push for us to use this sort of wording. As written it's:

- Vague: What counts as a SAST tool? There's basic linting, there's cyclomatic complexity checking, and all sorts of other stuff.
- Too strict for every change to a codebase: There are lots of changes to a code base that don't require static analysis, e.g. documentation changes

baseline.yaml
```diff
@@ -720,6 +720,55 @@ criteria:
...
+      with a cryptographic signature.    
+    objective: |
+      Provide users with a mechanism to verify the
+      authenticity and integrity of released 
+      software assets, reducing the risk of
+      tampering or unauthorized modifications.  
+    implementation: |
+      Sign all released software assets at build 
+      time with a cryptographic signature, such 
+      as a GPG or PGP signature.
```
TomHennen, line 739:
```suggestion
      time with a cryptographic signature or attestations, such 
      as GPG or PGP signature, Sigstore signatures, SLSA provenance, or SLSA VSAs.   
```

baseline.yaml
```diff
@@ -720,6 +720,55 @@ criteria:
...
+    objective: |
+      Provide users with a mechanism to verify the
+      authenticity and integrity of released 
+      software assets, reducing the risk of
+      tampering or unauthorized modifications.  
+    implementation: |
+      Sign all released software assets at build 
+      time with a cryptographic signature, such 
+      as a GPG or PGP signature.   
+
```
TomHennen, line 740:
```suggestion
      
      Provide clear documentation for users on how they can verify the signature
      and what a valid signature or attestation indicates (e.g.
      "it came from our CI/CD system",
      "one of our project maintainers built the software", "it was built from the
      indicated source repo with the indicated builder", "it meets SLSA level 3").
      
```

TomHennen: sorry, my suggestion may need some more formatting tweaks...

baseline.yaml
TomHennen: For PW 7.2 SSDF does call out the use of static analysis tools and is somewhat specific about how to handle the results "Example 4: Use a static analysis tool to automatically check code for vulnerabilities and compliance with the organization’s secure coding standards with a human reviewing the issues reported by the tool and remediating them as necessary."

I'd agree that we should probably call out specific examples of what these tools are.

eddie-knight: > Vague: What counts as a SAST tool? There's basic linting, there's cyclomatic complexity checking, and all sorts of other stuff. @mlieberman85 

Agreed that this needs to be defined. We have a lexicon definition of SCA, but currently lacking one for SAST.

> Too strict for every change to a codebase: There are lots of changes to a code base that don't require static analysis, e.g. documentation changes @mlieberman85 

Our lexicon segments the codebase from the documentation (We can also improve this if needed).

```yaml
  - term: Codebase
    definition: |
      The collection of source code and related
      assets that make up the project. The codebase
      includes all files necessary to build and
      test the software. Lives in the repository,
      sometimes alongside documentation and CI/CD
      pipelines. The contents of the codebase are
      the primary deliverable in a release.
```

TheFoxAtWork: I would pile on here (since i cannot find it in the criteria) that we lack the criteria to specific SAST or SCA must be conducted. It is assumed as criteria with the current osps-71 for SCA but not explicit. That criteria also specific policy for resolution of findings prior to release.  Here deviating from that, to focus more on the checks executed prior to merging PRs, which may or may not impact the release process (every project is different).

Given findings from either are derived from non-static data source content (new vulns constantly being reported), I would recommend exploring source code management and PR processes distinct from the build itself. If we're concerned about entities building from source, then forcing checks pass prior to merge makes sense. If we're more concerned about produced artifacts, then it should be as part of the build. if a project cuts releases off a particular branch, the reasonable expectation may be that those checks are conducted at merge to initiate cutting the release - in this case it is still met through the projects desired implementation.

baseline.yaml
eddie-knight: @mlieberman85 If the desire is for more than signing— should we split this into Lv2 signature, and Lv3 for more complete attestation?

baseline.yaml
```diff
@@ -720,6 +720,61 @@ criteria:
...
+    scorecard_probe:
+      - releasesAreSigned 
+  - id: OSPS-54
+    maturity_level: 3
+    category: Quality
+    criteria: |
+      All changes to the project's codebase MUST
+      require a passing status check from a
+      Static Application Security Testing (SAST)
+      tool.
```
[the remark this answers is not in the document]
eddie-knight: This accounts for the need of SAST. Recommend we create a SCA/dependency criteria in a follow-up PR.

baseline.yaml
mlieberman85: I think that's reasonable. I think we should also be clear what just blanket "signing" should be interpreted as. Since people sign a lot of stuff for a lot of reasons, which is why folks have been moving to attestations so it's clear what the intent of the signing is.

baseline.yaml
puerco: What about something like
```suggestion
      All released software assets MUST be signed or accounted for in a signed manifest including each asset's cryptographic hashes (such as an SBOM, provenance attestation or similar). The identity of the person or process authoring the software release must be published, discoverable and must be taken into account during the verification process.
```

baseline.yaml
puerco: I think the first level should be plain signing with published singer identities. It's way easier to implement plain signing and it gets us initial integrity guarantees.

baseline.yaml
```diff
@@ -720,6 +720,62 @@ criteria:
...
+      
+      It is recommended that this signature is
+      generated as part of the build and release
+      pipeline to ensure that it is consistent and
+      automated.
+    control_mappings: # TODO
+    security_insights_value: # TODO
+    scorecard_probe:
+      - releasesAreSigned 
+  - id: OSPS-54
```
funnelfiasco, line 755: I'd suggest splitting this out into a separate PR, since it seems less controversial than the other one.

puerco:
Following up on @funnelfiasco's comments I've opened https://github.com/ossf/security-baseline/pull/35 to separate the SCA check (vulnerability scan findings) into OSPS-55
/cc @SecurityCRob

eddie-knight:
Per suggestion from @funnelfiasco, we've extracted these suggestions into #89 and #91

PR #45, opened 2024-11-08
Allow passkeys

david-a-wheeler:
Passkey portability is being developed, and once developed, passkeys may become widespread.

Strictly speaking they may not meet the definition of MFA, but I think they should be permitted. They're way better than passwords. Let's make that clear.


eddie-knight:
Need input from others on whether passkey is an acceptable alternative to MFA.

david-a-wheeler:
I can see arguments either way. However, I think there's a very good chance that passkeys will become more common, so it needs to be clear.

It is *possible* to implement passkeys in a way that also meets MFA. However, the most likely approaches are client-side use of a password or biometric check to unlock use of the passkey private key. An organization could have a policy requiring this, but it's impractical to have any kind of automated check for that, and it's the sort of policy that would be trivially ignored.

SecurityCRob (approved):

baseline.yaml
```diff
@@ -53,13 +53,21 @@ criteria:
...
+      so depending solely on passwords is a weaker
+      form of authentication.
     implementation: |
-      Require multi-factor authentication for the
+      Require multi-factor authentication (MFA) for the
       project's version control system, requiring
       collaborators to provide a second form of
       authentication when accessing sensitive data
       or modifying repository settings.
+      For purposes of this criterion passkeys are considered
```
funnelfiasco, line 66: Minor reworking suggestion:

> For this criterion, passkeys are considered an

funnelfiasco (approved):
LGTM with a minor non-blocking suggestion.

If you end up making additional changes to this before it's merged, I'd suggest splitting the spelling corrections on lines 10 and 13 a separate commit.

puerco (approved):
Barring @funnelfiasco's nit above, I'm happy to see this addition.

PR #51, opened 2024-11-08
DCO/CLA changed to lv2

eddie-knight:


david-a-wheeler:
I like DCOs. That said, they definitely aren't done everywhere, and if the goal is make baseline-1 a real minimum, it's arguable this goes higher.

david-a-wheeler (approved):

SecurityCRob (approved):

PR #56, opened 2024-11-08
Removed requirments.txt from deps implementation

eddie-knight:


david-a-wheeler:
This is fine. Usually you don't need to duplicate "or" in a list, you could just say "A, B, or C". But it's not *wrong*, so carry on.

david-a-wheeler (approved):

baseline.yaml
```diff
@@ -637,8 +637,8 @@ criteria:
       This may take the form of a software bill of
       materials (SBOM) or a dependency file that
       lists all direct and transitive dependencies
-      such as requirements.txt, package.json, or
-      Gemfile.lock, or go.sum.
+      such as package.json, or Gemfile.lock, or
```
eddie-knight, line 640:
```suggestion
      such as package.json, Gemfile.lock, or
```

SecurityCRob (approved):

Issue #42, opened 2024-11-07
Formalize ID Format


eddie-knight:
A suggestion from @TheFoxAtWork is to consider incorporating the category into the ID.

> overall this looks fine and it makes sense to reorganize. Perhaps it may be beneficial to alter the numbering scheme to reflect the category?
> OSPS-AC-1.1
> OSPS-AC-1.2
> OSPS-AC-1.3
> OSPS-AC-2.1
> OSPS-AC-2.2
> 
> in doing this method, it allows for sorting by category. the numerical value also indicates the level and component under that level we're referring to. further, we can continue to add criteria for any we miss without having to rework criteria numbering and reorganization. this means in the event new categories need added, we have a schema for them.
> 
> https://github.com/ossf/security-baseline/pull/34#pullrequestreview-2408680153

david-a-wheeler:
I recommend NOT using numbers for ids at *ALL*. Instead, use short names. Numbers have no intrinsic meaning, and using number ranges will mess up everything if a category overflows. Using simple names makes everything MUCH MUCH easier to follow.

E.g., instead of `OSPS-01` use `OSPS-MFA` or similar.

TheFoxAtWork:
Also perfectly acceptable so long as we don't have to recategorize these if we need to add or remove them.

eddie-knight:
@david-a-wheeler and @SecurityCRob have agreed to suggest some alternative formats.

funnelfiasco:
Another reason to go with descriptive IDs instead of numerical: we've changed the level of some requirements without renumbering them (e.g. #51). So if we don't switch to descriptive IDs (which I think is probably better, but may be challenging), we may just want to have the numerical IDs be sequential like RFCs or PEPs.

drusso-rh:
(Redirected from PR 72)
FWIW I agree that there is value in having a unique ID for each requirement. Many frameworks and guidance docs do this, and it is invaluable when mapping/cross-walking these artifacts with an organization's internal requirements.

funnelfiasco:
I’ve thought about this some and I’ve landed on the position that we should use numeric, sequential identifiers here.

I understand — and agree with — the argument that numerical IDs don’t tell the reader anything. On the other hand, coming up with short, descriptive IDs can be challenging too. For example: OSPS-03 and OSPS-04. Those are similar enough that the descriptive IDs would either have to be not very short or so compressed that they’re no longer much more descriptive.

[...]

PR #72, opened 2024-11-14
Remove id range note

david-a-wheeler:


eddie-knight (approved):
Yes, this is no longer true as we've changed around the maturity levels.

This does not address the need for a new ID pattern.

SecurityCRob:
The ranges, as listed, are no longer valid today.  Can we create new ranges and adjust the identifiers accordingly?  We have 5 high-level categories, could each group be placed in a range of 20 #s, giving us the option of 100 total criteria?  Today Badges has ~137 criteria, if we pad for future requirements, we could use sets of 30 ids for each category, giving us room to match BPB and have room to expand further.

eddie-knight:
Considering that we're still pending an ID redesign, I think this is mostly a moot point. Removing the concept of a reserved range is quite fine

david-a-wheeler:
I think using numbers as IDs is a bad idea for requirements. Numbers have no meaning. If you make a mistake in a mapping, you typically won't notice. If all you have is the id, you have no idea what it's about. If you delete one, everyone asks why there's a missing one, which isn't the point.

In contrast, if you see requirement "OSPS-MFA" you can probably guess what it means.

eddie-knight:
@SecurityCRob This one should be good to merge, as we're all in agreement that the ID pattern is going to be redesigned

funnelfiasco (approved):

SecurityCRob:
> I think using numbers as IDs is a bad idea for requirements. Numbers have no meaning. If you make a mistake in a mapping, you typically won't notice. If all you have is the id, you have no idea what it's about. If you delete one, everyone asks why there's a missing one, which isn't the point.
> 
> In contrast, if you see requirement "OSPS-MFA" you can probably guess what it means.

STRONGLY disagree about the complete removal of numbering.  800-53, NIST SSDF, CSF, BSI's CRA Guidance, etc. all use alpha-numeric identifiers in their schemes.  Shortening the names of 50ish requirements does not yield a meaningful identifier.  Someone may have no context or background on what "MFA" is, but is able to grok "2FA" or "Mutifactor" better?

eddie-knight:
Since the conversation has drifted into overlap with #42, I'm going to close this PR. We can figure out how to replace the current numbering format in that issue, or in a future working session.

PR #84, opened 2024-11-21
Modify 'direct' to deal with decentralized repos

david-a-wheeler:
The current baseline text unintentionally presumes that a VCS must work centrally. That is, it assumes that contributors often share write access to the shared repo.

Many projects (like the Linux kernel) do NOT operate this way. In more decentralized models, changes are proposed in a separate repo. Merging those proposed changes is a separate request to the main repo (and its primary branch), which may or may not be granted. This still counters unintentional changes, and meets the underlying goal, but in a different way
than the baseline text currently requires.

This attempts to modify the baseline so that it no longer


SecurityCRob (approved):

funnelfiasco (approved):
This is a good improvement to make this criterion more broadly-applicable.

david-a-wheeler:
There was a discussion on if this should be level 1, that can be a different PR.

puerco (approved):

mlieberman85 (approved):

SecurityCRob:
from our 1/7/25 meeting:
- Desire to avoid straight merges and have some person or automation review all changes
- Group is ok with intent, will discuss level later
- Evan will open a PR to (discuss) changing the target maturity level

PR #90, opened 2024-11-27
Remove the lexicon file

funnelfiasco:
This file appears to be unused. The lexicon comes from the yml file.

@eddie-knight is my understanding correct?


eddie-knight (approved):
Yep

puerco (approved):

PR #92, opened 2024-11-30
Added basic field validation

### LINKED ISSUE #88: [bug] build script does not check for invalid or duplicate IDs
The build script does not currently check to see if an ID appears twice or if an ID does not follow our defined format, which means we could end up re-using an ID by accident or adding an invalid ID.

I think the best way would be to add a `check` command that validates IDs (and possibly other fields), so that we can run just the `check` as a CI workflow on pull requests. The `compile` command, which produces actual output, could either call the check first, or we can adjust the pages workflow to include a validate step before the compile step as a backstop against something sneaking in accidentally.

eddie-knight:
Resolves #88

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -111,7 +111,7 @@ func readYAMLFile() error {
 	var entryIDs []string
 	for i, entry := range baseline.Criteria {
 		// if entry in entryIDs
-		if contains(entryIDs, entry.ID) {
+		if slices.Contains(entryIDs, entry.ID) {
 			return fmt.Errorf("duplicate ID for criteria entry %d: %s", i, entry.ID)
 		}
 		if entry.ID == "" {
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -108,8 +108,37 @@ func readYAMLFile() error {
 	if err := decoder.Decode(&baseline); err != nil {
 		return fmt.Errorf("error decoding YAML: %v", err)
 	}
+	var entryIDs []string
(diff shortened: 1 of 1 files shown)


funnelfiasco (approved):
LGTM. We'll still want to validate that the ID follows the specification, but that can be a separate change.

cmd/main.go
```diff
@@ -108,10 +108,39 @@ func readYAMLFile() error {
 	if err := decoder.Decode(&baseline); err != nil {
 		return fmt.Errorf("error decoding YAML: %v", err)
 	}
+	var entryIDs []string
+	for i, entry := range baseline.Criteria {
+		// if entry in entryIDs
+		if contains(entryIDs, entry.ID) {
```
puerco, line 114: Prefer the native method here (and drop the custom func on line 135)
```suggestion
		if slices.Contains(entryIDs, entry.ID) {
```

cmd/main.go
```diff
@@ -108,10 +108,39 @@ func readYAMLFile() error {
...
+			return fmt.Errorf("duplicate ID for criteria entry %d: %s", i, entry.ID)
+		}
+		if entry.ID == "" {
```
puerco, line 117: Now that we have a naming pattern, we should add a regexp check here (we can follow up with that later)

puerco: Ah I just saw this is the same comment @funnelfiasco made :)

SecurityCRob (approved):

Issue #94, opened 2024-12-02
Sort entries according to category


funnelfiasco:
Mentioned by @eddie-knight in [issue 88](https://github.com/ossf/security-baseline/issues/88#issuecomment-2508663965):

> we need to sort the entries according to category now that https://github.com/ossf/security-baseline/pull/87 is merged! Puerco and I noticed that it takes a bit of searching to make sure you're using the right ID when making a new entry.

funnelfiasco:
@eddie-knight did you mean in the file or in the output? I'm assuming in the file, but before I tackle this I want to make sure.

eddie-knight:
Hey @funnelfiasco — I was working on this, but reviewing for typos got my eyes heavy 😆 Just opened the PR and linked it here

funnelfiasco:
This was fixed in #95

PR #95, opened 2024-12-02
Reorganize

eddie-knight:
Addresses #94


funnelfiasco (approved):
Dang, you're fast! Looks good to me.

puerco (approved):
There a re a few nits I'd like to address but let's get this in as its a big change and keep iterating. Thanks Eddie!

SecurityCRob (approved):

PR #98, opened 2024-12-03
go mod tidy + ci check for site build

eddie-knight:


.github/workflows/test-build.yml
```diff
@@ -0,0 +1,23 @@
...
+permissions:
+  contents: read
+
+jobs:
+  # Build job
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
```
funnelfiasco, line 16: Let's have this action drop the credentials after it's done by adding

```
        with:
            persist-credentials: false
```

See https://github.com/ossf/security-baseline/blob/main/.github/workflows/web-publish.yml#L27-L30

funnelfiasco (approved):
LGTM. This is a good addition. I think you should make the suggested change re: persistent credentials, but I won't block on it.

It would be good as a next step to have some way to view the rendered output so that we can find more subtle issues, but this will at least catch the glaring ones.

.github/workflows/test-build.yml
eddie-knight:
```suggestion
        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
        with:
            persist-credentials: false
```

Like so?

funnelfiasco: #nailedit

SecurityCRob (approved):

Issue #99, opened 2024-12-03
Proposal: Human-readable format should not be a spreadsheet or table-style document


jordansissel:
As the title suggests, I propose that the human readable format for this baseline stuff should be something *not* a table or spreadsheet. The long-form text of exposition and explanation should not be typeset into a skinny spreadsheet-like structure. For a visual example, this project's current website looks like this on my phone:

![image](https://github.com/user-attachments/assets/088875b0-d77d-490c-9daf-eda1076929ca)

Too much whitespace is dedicated to the data columns with the least amount of text.  Even when I scroll across, the table permits at most 3 words per line. It's not great ;) Yes, it looks different on my computer with a bigger screen and different layout provided by the browser on this table.

The longer "criteria details" area below the table feels informative enough without needing the table above it.

Thoughts?

funnelfiasco:
The current styling of the webpage was not designed with mobile browsing in mind (and it wasn't even checked on a mobile browser), but that doesn't mean it can't be. If the small-window view gave the last column more space, would that be sufficient to address your concerns?

If not, do you have any suggestions for an alternative presentation? We want the table (or whatever replaces it) to be relatively dense, so that people can get a quick understanding of the criteria and then click through to the one they want more details on.

What about something like this?

```
OSPS-EX-01  1  Example
     Examples MUST NOT use lorem ipsum because Latin gibberish is still gibberish.
OSPS-EX-02  1  Example
     Examples must use domain names reserved for that purpose as described in section 6.5 of RFC 6761
```

david-a-wheeler:
I think what you want is a "responsive table" - that is, the formatting radically changes when you have a narrow screen so it's still readable by breaking it differently. Unsuprisingly, there are many ways to do this. A few links:
https://css-tricks.com/responsive-data-table-roundup/
https://codepen.io/shellbryson/post/responsive-tables
https://github.com/rootwork/responsive-tables-builder (links to some options)

We should look at options, and then apply it.

eddie-knight:
@SecurityCRob we do have spots in the current template (and yaml) for control mappings!

eddie-knight:
Happy to update the template with a change to the top summary section.

Should we...

🎉 - Change to an abbreviated list, instead of table
🚀 - Try one of the responsive table approaches
👀 - Think up another idea

funnelfiasco:
[...]

PR #103, opened 2024-12-10
Clarify external workflow. Fixes #63

### LINKED ISSUE #63: Clarify code outside build script
This criterion is *not* clear:

~~~~
  - id: OSPS-05
    maturity_level: 1
    category: Build & Release
    criteria: |
      The project's build and release pipelines
      MUST NOT execute arbitrary code that is
      input from outside of the build script.
    objective: |
      the execution of external code.
    implementation: |
      pipelines do not execute arbitrary code
      provided from external sources.
~~~~

On first reading, it sounded like many common practices
are forbidden:

* pipe-to-shell like `curl https://STUFF | sh` appears forbidden.
  I'm no fan of pipe-to-shell, as it carries big risks if the
  server is compromised, but it's so widespread that it can't be
  a level 1 requirement. For example, the recommended way to
  install Rust is pipe-to-shell.
* Downloading components & dependencies to run the
  build script appears forbidden - they must all be downloaded
  before building. Again, not a bad idea, but *way* too many
  build systems figure out what's needed, then get it.

It appears what's really meant is countering some specific
workflow attacks like script injection, as detected by Scorecard.
That sounds far more tractable, but the current text doesn't *say* that.

I think this text needs to be redone to say what was actually intended.

david-a-wheeler:
The original requirement is unclear. The discussion on 2024-12-10 tried to clarify what was intended.
This is one attempt to clarify things.

### DIFF
--- a/baseline.yaml
+++ b/baseline.yaml
@@ -129,18 +129,20 @@ criteria:
     maturity_level: 1
     category: Build & Release
     criteria: |
-      The project's build and release pipelines
-      MUST NOT execute arbitrary code that is
-      input from outside of the build script.
+      A project’s build process that either leads to release
+      or can read secret data MUST NOT execute untrusted external code.
     objective: |
       Reduce the risk of code injection or other
       security vulnerabilities in the project's
       build and release processes by restricting
-      the execution of external code.
+      the execution of external code in workflows.
     implementation: |
       Ensure that the project's build and release
-      pipelines do not execute arbitrary code
+      pipelines do not execute untrusted code
       provided from external sources.
+      One approach is to only allow maintainers to identify
+      which external code can be used, using verification mechanisms
+      such as digital signatures or https.
     control_mappings: # TODO
     scorecard_probe:
       - hasDangerousWorkflowScriptInjection
--- a/baseline.yaml
+++ b/baseline.yaml
@@ -129,8 +129,8 @@ criteria:
     maturity_level: 1
     category: Build & Release
     criteria: |
-      A project’s build process that either leads to release
-      or can read secret data MUST NOT execute untrusted external code.
+      A build process that either leads to release or can read secret data
+      MUST NOT execute untrusted code.
     objective: |
       Reduce the risk of code injection or other
       security vulnerabilities in the project's


baseline.yaml
```diff
@@ -129,18 +129,20 @@ criteria:
     maturity_level: 1
     category: Build & Release
     criteria: |
-      The project's build and release pipelines
-      MUST NOT execute arbitrary code that is
-      input from outside of the build script.
+      A project’s build process that either leads to release
+      or can read secret data MUST NOT execute untrusted external code.
```
funnelfiasco, line 133: In the interests of brevity, I'd rewrite this as:

```suggestion
      A build process that either leads to release or can read secret data
      MUST NOT execute untrusted code.
```
The more we can excise unnecessary words from the criteria, the easier it will be for people to quickly read, addressing part of the concern in #99. I can see an argument for leaving "external" in, but I'd counter that we don't want the build pipeline running untrusted external code either, and the objective & implementation fields give us more space to expand on what we're trying to say.

That said, I'm not sure I like explicitly calling out "read secret data" here. Are there other times that untrusted code in the build process is of concern?

funnelfiasco (commented):
I don't have significant objections to this, but I'll hold off on adding approval until I submit my version and then we can let the WG compare them and see how we should best blend them.

david-a-wheeler:
@funnelfiasco - here's my try!

david-a-wheeler:
I accepted the change. I'm a little worried that removing the word "external" removes too much, since the objective is really about external code, but maybe this is okay. It's at least clearer than it was before.

baseline.yaml
```diff
@@ -129,18 +129,20 @@ criteria:
     maturity_level: 1
     category: Build & Release
     criteria: |
-      The project's build and release pipelines
-      MUST NOT execute arbitrary code that is
-      input from outside of the build script.
+      A build process that either leads to release or can read secret data
+      MUST NOT execute untrusted code.
```
eddie-knight, line 133: We'd also need to cover any build pipelines that produce artifacts that would be used on a system that might contain sensitive information, such as an end user or developer machine. 🤔 Effectively any pipeline that produces any artifact that could be used by any machine that isn't ephemeral.

david-a-wheeler: Would this work?: s/secret/sensitive/

eddie-knight:
Closing this and #104 together in favor of a follow-up to either remove this or improve the language to align with the intent behind Scorecard's `hasDangerousWorkflowScriptInjection`

david-a-wheeler:
In later discussions, we found that the real goal is to capture Scorecard's requirement, but that's not obvious. Let's try again....

PR #104, opened 2024-12-10
Clarify OSPS-BR-01 to better express the intent

funnelfiasco:
This is my own attempt at #63. It differs from @david-a-wheeler's #103 by:

* Keeping the criterion inclusive of build pipelines. 
* Explicitly calling out that we do not mean to prohibit the use of RPM, deb, etc.

It does include some edits from David's PR as well.

### DIFF
--- a/baseline.yaml
+++ b/baseline.yaml
@@ -197,7 +197,7 @@ criteria:
     category: Build & Release
     criteria: |
       Build and release pipelines MUST only execute
-      code that's present in the repository *or* from external trusted sources.
+      code that's either present in the repository *or* from external sources trusted by the project.
     objective: |
       Reduce the risk of code injection or other
       security vulnerabilities in the project's
--- a/baseline.yaml
+++ b/baseline.yaml
@@ -196,8 +196,8 @@ criteria:
     maturity_level: 1
     category: Build & Release
     criteria: |
-      Build and release pipelines MUST NOT execute
-      untrusted code.
+      Build and release pipelines MUST only execute
+      code that's present in the repository *or* from external trusted sources.
     objective: |
       Reduce the risk of code injection or other
       security vulnerabilities in the project's
--- a/baseline.yaml
+++ b/baseline.yaml
@@ -196,18 +196,24 @@ criteria:
     maturity_level: 1
     category: Build & Release
     criteria: |
-      The project's build and release pipelines
-      MUST NOT execute arbitrary code that is
-      input from outside of the build script.
+      Build and release pipelines MUST NOT execute
+      untrusted code.
     objective: |
       Reduce the risk of code injection or other
       security vulnerabilities in the project's
       build and release processes by restricting
-      the execution of external code.
+      the execution of external code in workflows.
     implementation: |
       Ensure that the project's build and release
-      pipelines do not execute arbitrary code
-      provided from external sources.
+      pipelines do not execute untrusted code
+      provided from external sources. Maintainers
+      may establish trust in a variety of ways,
+      including digital signature verification and
+      inspection of external code. For clarity,
+      this criterion does not prohibit the use of
+      software in a package format that executes
+      scripts (e.g. RPM, .deb) so long as the
+      package itself is trusted.
     control_mappings: # TODO
     scorecard_probe:
       - hasDangerousWorkflowScriptInjection


funnelfiasco:
Commentary on the "build and release" versus "release + build in certain cases":

As I think about it, beyond the concern with "does 'release and can read secret data' sufficiently cover all of the things we're trying to guard against?" that I raised in my comment on #103, I am also worried about the cognitive load that a more narrowly-carved criterion might create.

I'm sympathetic to the argument David made in today's meeting about having a solid justification for adding a criterion. In this particular case, though, I worry that trying to figure out exactly where the boundary is will distract maintainers from taking the action necessary to secure their workflows. As noted on the call by someone, most builds don't get released, but my hunch is that most build _workflows_ could lead to a release, so there's not a lot of benefit in saying "if you have a build workflow that you know will never lead to a release, you can run whatever you want on it and we don't care". Or at least the benefit of not caring doesn't outweigh the added complexity.

baseline.yaml
```diff
@@ -196,18 +196,24 @@ criteria:
...
-      the execution of external code.
+      the execution of external code in workflows.
     implementation: |
       Ensure that the project's build and release
-      pipelines do not execute arbitrary code
-      provided from external sources.
+      pipelines do not execute untrusted code
+      provided from external sources. Maintainers
+      may establish trust in a variety of ways,
+      including digital signature verification and
```
funnelfiasco, line 211: I don't entirely love my wording here. I think the intent is right, but it could use some polish.

SecurityCRob: i think this is good enough for now, i a revision idea comes up before this merges I'll try to suggest it.

SecurityCRob (approved):

baseline.yaml
```diff
@@ -196,18 +196,24 @@ criteria:
...
+      inspection of external code. For clarity,
+      this criterion does not prohibit the use of
+      software in a package format that executes
+      scripts (e.g. RPM, .deb) so long as the
+      package itself is trusted.
```
evankanderson, line 216: I'm struggling with how to enforce this.  For example, does this prohibit using a tool built and maintained in the same repository which automates some of the release steps unless we can ensure that the tool cannot be subverted?  As written, it suggests every Makefile and data processor needs a security review, which seems high for maturity level 1.

Similarly, it's not clear whether this criteria establishes a transitive duty of care to audit not just your own codebase, but those of pipeline dependencies (e.g., do I need to check on untrusted code execution from `actions/setup-go` in my GitHub workflows)?  Or can I say "I'm using this action from $ELSEWHERE, I assume it's secure" at level 1, and have additional security criteria at level 2 or 3?

baseline.yaml
```diff
@@ -196,18 +196,24 @@ criteria:
...
-      MUST NOT execute arbitrary code that is
-      input from outside of the build script.
+      Build and release pipelines MUST NOT execute
+      untrusted code.
     objective: |
       Reduce the risk of code injection or other
       security vulnerabilities in the project's
       build and release processes by restricting
-      the execution of external code.
+      the execution of external code in workflows.
```
evankanderson, line 205: I'm not sure what "in workflows" here is supposed to capture compared with the older version.

evankanderson (commented):
I've been struggling a bit with how to validate this criteria (as writ, it seems to require avoiding security bugs in writing release software -- I assume this is aspirational, but it doesn't provide a clear "you are done" threshold for either project authors or validation software).

baseline.yaml
funnelfiasco: The idea is to scope it specifically to build and release workflows as opposed to, say, the user execution of the program. Since it's in the "Build and Release" category, I'd accept an argument that it is redundant.

baseline.yaml
funnelfiasco: It does not establish a transitive duty of care. As an example, a workflow that uses an Ubuntu image from Canonical wouldn't violate this because it's a trusted source. A workflow that uses `some-random-rebuild/ubuntu-latest` would, unless the project has established trust in the `some-random-rebuild` project. The idea is more along the lines of "we're getting the code we think we're getting" as opposed to "we have inspected every line of software all the way down and it is clean".

For the first part, I would say if it's in the same repository, it's part of the project and therefore not an "external source".

david-a-wheeler: > As an example, a workflow that uses an Ubuntu image from Canonical wouldn't violate this because it's a trusted source.

But it _would_ violate this requirement as written. The term "trusted source" isn't present in this text.

I'll try to make a change, I think the point is that build only runs code in the repo *or* from external trusted sources.

baseline.yaml
```diff
@@ -196,18 +196,24 @@ criteria:
     maturity_level: 1
     category: Build & Release
     criteria: |
-      The project's build and release pipelines
-      MUST NOT execute arbitrary code that is
-      input from outside of the build script.
+      Build and release pipelines MUST NOT execute
+      untrusted code.
```
david-a-wheeler, line 200:
```suggestion
      external code untrusted code.
```

david-a-wheeler, line 200:
```suggestion
      Build and release pipelines MUST only execute
      code that's present in the repository *or* from external trusted sources.
```

baseline.yaml
```diff
@@ -196,18 +196,24 @@ criteria:
     maturity_level: 1
     category: Build & Release
     criteria: |
-      The project's build and release pipelines
-      MUST NOT execute arbitrary code that is
-      input from outside of the build script.
+      Build and release pipelines MUST only execute
+      code that's present in the repository *or* from external trusted sources.
```
david-a-wheeler, line 200:
```suggestion
      code that's either present in the repository *or* from external sources trusted by the project.
```

The text said "trusted" but not "trusted by whom?" We need to make it clear that it's the *project* that determines what external sources are trusted. I put the "either" there to make sure it's clear the *or* separates two clauses.

evankanderson:
Additionally, the scorecard check is more about workflows that run on PR, rather than on project build.

I think both are important, but the reference to the scorecard check here is confusing.

eddie-knight:
Closing this and #104 together in favor of a follow-up to either remove this or improve the language to align with the intent behind Scorecard's `hasDangerousWorkflowScriptInjection`

PR #109, opened 2024-12-15
fix: change IDs to sequence with maturity levels

eddie-knight:


jmeridth (approved):

funnelfiasco (approved):

funnelfiasco (commented):
I originally marked this as approved, but now I'm reconsidering. Is there value in making this change, since subsequent modifications/additions/removals will almost certainly make the IDs non-sequential anyway?

baseline.yaml
```diff
@@ -766,6 +740,32 @@ criteria:
     scorecard_probe:
       - # None, may need to be paired with SI
 
+  - id: OSPS-LE-04
```
SecurityCRob, line 743: this already is le-01 in the current yaml.  le-04 and le-02 are dupes, so le-04 should be removed

baseline.yaml
```diff
@@ -960,6 +927,39 @@ criteria:
     security_insights_value: # TODO
     scorecard_probe: # sastToolRunsOnAllCommits 
 
+  - id: OSPS-QA-07
```
SecurityCRob, line 930: qa-07 was introduced earlier, so old qa-05 can not move here.  Is theidea simply to move qa-05 from lvl 2 to lvl 3?  If so, can we just keep it the same id and adjust the level to 3?

SecurityCRob:
> I originally marked this as approved, but now I'm reconsidering. Is there value in making this change, since subsequent modifications/additions/removals will almost certainly make the IDs non-sequential anyway?

We're going to running into this until we get the criteria locked down, and even after that any future criteria added will probably be "out of level sequence".  I would suggest, for now, to NOT renumber/reorder.  I have a few additional criteria for the group to consider to add too based on other frameworks/regs which will mess any order we decide today up.

david-a-wheeler:
I think trying to change IDs to be sequential is not a good idea. If it's used, it will be maintained, and if it's maintained, the numbers will necessarily stop being a sequence.

That's why I recommended using names, and not numbers, in the first place.

If you must use numbers, make it clear that they are simply unique IDs. Don't bother trying to make them sequential. Trying to make them sequential creates a lot of extra unnecessary work (especially for people creating mappings) and it will always fail sooner or later.

PR #113, opened 2024-12-17
Present criteria as not-tables

### LINKED ISSUE #99: Proposal: Human-readable format should not be a spreadsheet or table-style document

funnelfiasco:
This improves small-screen readability and removes/moves extraneous information for easier skimming.

Some choices I made:

* I went with just the ID and criterion because the level is in the heading and the category is in the ID. I added them to the detailed view for completeness sake.
* I could have iterated over the three levels instead of copy pasta, but
    1. The more code blocks we add, the more complicated the template gets
    2. I think we'll eventually want to add a brief description of each level before we start listing the criteria, so this is future-proofing against that a bit
* Edit to add the other thing I was going to put here: this approach requires no muddling with CSS, adding plugins, etc and still achieves the overarching goal

Preview available as https://funnelfiasco.github.io/security-baseline/

Fixes #99

### DIFF
--- a/cmd/template.md
+++ b/cmd/template.md
@@ -15,11 +15,32 @@ For more information on the project and to make contributions, visit the [GitHub
 
 ## Criteria Overview
 
-| ID  | Maturity Level | Category | Criteria |
-| --- | -------------- | -------- | -------- |
+* [Level 1](#level-1)
+* [Level 2](#level-2)
+* [Level 3](#level-3)
+
+### Level 1
 
 {{- range .Criteria }}
-| [{{ .ID }}]({{ .ID | asLink }}) | {{ .MaturityLevel }} | {{ .Category }} | {{ .CriteriaText | collapseNewlines | addLinks }} |
+{{if eq .MaturityLevel 1}}
+**[{{ .ID }}]({{ .ID | asLink }})**: {{ .CriteriaText | addLinks }}
+{{ end }}
+{{- end }}
+
+### Level 2
+
+{{- range .Criteria }}
+{{if eq .MaturityLevel 2}}
+**[{{ .ID }}]({{ .ID | asLink }})**: {{ .CriteriaText | addLinks }}
+{{ end }}
+{{- end }}
+
+### Level 3
+
+{{- range .Criteria }}
+{{if eq .MaturityLevel 3}}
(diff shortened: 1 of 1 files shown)


eddie-knight:
Thanks for making this happen, I think this is the right approach.

Something feels just a bit off with the list, but I can't put my finger on it. Perhaps it's the lack of a leading list mark on each line, or the use of an emdash 🤔 Just feels like it's _almost_ there, but not quite.

jordansissel:
This is much improved when I try to read this version compared to the table (especially on mobile)

cmd/template.md
```diff
@@ -15,11 +15,32 @@ For more information on the project and to make contributions, visit the [GitHub
...
-| --- | -------------- | -------- | -------- |
+* [Level 1](#level-1)
+* [Level 2](#level-2)
+* [Level 3](#level-3)
+
+### Level 1
+
+{{- range .Criteria }}
+{{if eq .MaturityLevel 1}}
+[{{ .ID }}]({{ .ID | asLink }}) — {{ .CriteriaText | addLinks }}
```
eddie-knight, line 26:
```suggestion
**[{{ .ID }}]({{ .ID | asLink }})** - {{ .CriteriaText | addLinks }}
```

david-a-wheeler:
This is an improvement. I suggest we merge something soon, and then we can continue to refine it further as later PRs. Making it readable at all on small displays seems important to me.

david-a-wheeler (approved):

funnelfiasco:
It made me very sad to remove my emdashes, but I did.

eddie-knight (approved):
Thanks for doing this @funnelfiasco!

PR #115, opened 2024-12-17
Add criterion for support statements - NEW - OSPS-DO-13

SecurityCRob:
new proposed criteria around support statements

### DIFF
--- a/baseline.yaml
+++ b/baseline.yaml
@@ -663,8 +663,9 @@ criteria:
        will be provided.
     implementation: |
        Create a status check that checks the project's
-       version control system for support and/or lifecycle
-       statements or machine-readable metadata.
+       publicly-available collateral for support 
+       and/or lifecycle statements or machine-readable 
+       metadata.
     control_mappings: # TODO
     security_insights_value: # TODO
     scorecard_probe: #  
--- a/baseline.yaml
+++ b/baseline.yaml
@@ -653,7 +653,7 @@ criteria:
     maturity_level: 1
     category: Documentation
     criteria: |
-       The PROJECT DOCUMENTATION MUST include a 
+       The project documentation MUST include a 
        descriptive statement about how the project 
        is supported/maintained and what is not supported. 
        Projects are encouraged to provide "End of 
--- a/baseline.yaml
+++ b/baseline.yaml
@@ -662,10 +662,11 @@ criteria:
        or may not provide and when no further updates
        will be provided.
     implementation: |
-       Create a status check that checks the project's
-       publicly-available collateral for support 
-       and/or lifecycle statements or machine-readable 
-       metadata.
+       Provide text identifying which versions/branches of the project
+       are being supported and at what level (e.g., improvements welcome,
+       bug fixes only, security fixes only, or abandoned).
+       Where practical, provide lifecycle statements or machine-readable 
+       metadata, e.g., via security-insights.yml project-lifecycle.status.       
     control_mappings: # TODO
     security_insights_value: # TODO
     scorecard_probe: #  
--- a/baseline.yaml
+++ b/baseline.yaml
@@ -656,9 +656,6 @@ criteria:
        The project documentation MUST include a 
        descriptive statement about how the project 
        is supported/maintained and what is not supported. 
-       Projects should provide "End of 
-       Support" statements when versions or features 
-       are deprecated or no longer supported.
     objective: |
        Identify and set expectations for downstream 
        consumers around what updates the project may
--- a/baseline.yaml
+++ b/baseline.yaml
@@ -656,7 +656,7 @@ criteria:
        The project documentation MUST include a 
        descriptive statement about how the project 
        is supported/maintained and what is not supported. 
-       Projects are encouraged to provide "End of 
+       Projects should provide "End of 
        Support" statements when versions or features 
        are deprecated or no longer supported.
     objective: |
--- a/baseline.yaml
+++ b/baseline.yaml
@@ -649,6 +649,29 @@ criteria:
     security_insights_value: # TODO
     scorecard_probe: # TODO
 
+  - id: OSPS-DO-13
+    maturity_level: 1
+    category: Documentation
+    criteria: |
+       The PROJECT DOCUMENTATION MUST include a 
+       descriptive statement about how the project 
+       is supported/maintained and what is not supported. 
+       Projects are encouraged to provide "End of 
+       Support" statements when versions or features 
+       are deprecated or no longer supported.
+    objective: |
+       Identify and set expectations for downstream 
+       consumers around what updates the project may
+       or may not provide and when no further updates
+       will be provided.
+    implementation: |
+       Create a status check that checks the project's
+       version control system for support and/or lifecycle
+       statements.
+    control_mappings: # TODO
+    security_insights_value: # TODO
+    scorecard_probe: #  
+
   - id: OSPS-LE-01
     maturity_level: 2
     category: Legal
--- a/baseline.yaml
+++ b/baseline.yaml
@@ -667,7 +667,7 @@ criteria:
     implementation: |
        Create a status check that checks the project's
        version control system for support and/or lifecycle
-       statements.
+       statements or machine-readable metadata.
     control_mappings: # TODO
     security_insights_value: # TODO
     scorecard_probe: #  
--- a/baseline.yaml
+++ b/baseline.yaml
@@ -654,8 +654,8 @@ criteria:
     category: Documentation
     criteria: |
        The project documentation MUST include a 
-       descriptive statement about how the project 
-       is supported/maintained and what is not supported. 
+       descriptive statement about the scope and 
+       duration of support. 
     objective: |
        Identify and set expectations for downstream 
        consumers around what updates the project may


baseline.yaml
```diff
@@ -649,6 +649,29 @@ criteria:
     security_insights_value: # TODO
     scorecard_probe: # TODO
 
+  - id: OSPS-DO-13
+    maturity_level: 1
+    category: Documentation
+    criteria: |
+       The PROJECT DOCUMENTATION MUST include a
```
eddie-knight, line 656:
```suggestion
       The project documentation MUST include a 
```

baseline.yaml
```diff
@@ -649,6 +649,29 @@ criteria:
...
+  - id: OSPS-DO-13
+    maturity_level: 1
+    category: Documentation
+    criteria: |
+       The PROJECT DOCUMENTATION MUST include a 
+       descriptive statement about how the project 
+       is supported/maintained and what is not supported. 
+       Projects are encouraged to provide "End of 
+       Support" statements when versions or features 
+       are deprecated or no longer supported.
```
eddie-knight, line 661: Encouragements aren't controls. This needs rephrased or moved to the objective / implementation.

baseline.yaml
```diff
@@ -649,6 +649,29 @@ criteria:
...
+    objective: |
+       Identify and set expectations for downstream 
+       consumers around what updates the project may
+       or may not provide and when no further updates
+       will be provided.
+    implementation: |
+       Create a status check that checks the project's
+       version control system for support and/or lifecycle
+       statements.
```
eddie-knight, line 670:
```suggestion
       Create a status check that checks the project's
       version control system for support statements, lifecycle
       statements, or related machine-readable metadata.
```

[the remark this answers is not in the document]
eddie-knight: Added a revised suggestion below for grammar/etc

funnelfiasco, line 670:
```suggestion
       Create a status check that checks the project's
       version control system, published documentation,
       or third-party service for support and/or lifecycle
       statements, or related machine-readable metadata.
```

funnelfiasco (changes requested):
This is pretty good overall, but there are a few changes I'd like to see before merging.

baseline.yaml
SecurityCRob: great suggestion, I like both!

eddie-knight: I think this is a good thing to discuss with Security Insights in mind. Currently, the design of SI is to have one insights file per repo, and [lifecycle](https://github.com/ossf/security-insights-spec/blob/main/specification/project-lifecycle.md) is included as (IIRC) an optional value.

Obligatory note: we have yet to revise security insights.

baseline.yaml
SecurityCRob: I'd LIKE to make it a requirement, but understand that not all projects send out EOL notices.  Should this be adjust to be a more directive statement or split out to a higher level criteria by itself?

funnelfiasco: I'd support sending EOL notices as a level 3 (arguably 2, but I'd like to start with 3 and be talked into 2) requirement.

baseline.yaml
funnelfiasco: I don't see anything in my edit that would preclude the use of Security Insights, but I don't think we want to have Baseline be tightly coupled to it.

baseline.yaml
SecurityCRob: OK, how about we split out that part to lvl and leave the broader "provide support statement" for lvl1?

funnelfiasco: Works for me.

SecurityCRob: adjusted and will propose additional criteria for EoS statement

baseline.yaml
```diff
@@ -649,6 +649,26 @@ criteria:
...
+       is supported/maintained and what is not supported. 
+    objective: |
+       Identify and set expectations for downstream 
+       consumers around what updates the project may
+       or may not provide and when no further updates
+       will be provided.
+    implementation: |
+       Create a status check that checks the project's
+       version control system for support and/or lifecycle
+       statements or machine-readable metadata.
```
funnelfiasco, line 667: I still think we should explicitly allow for other methods, but so long as the implementation is a "Pirates's Code", then I won't block on this.

baseline.yaml
```diff
@@ -649,6 +649,26 @@ criteria:
     security_insights_value: # TODO
     scorecard_probe: # TODO
 
+  - id: OSPS-DO-13
+    maturity_level: 1
+    category: Documentation
+    criteria: |
+       The project documentation MUST include a 
+       descriptive statement about how the project 
+       is supported/maintained and what is not supported.
```
funnelfiasco, line 658:
```suggestion
       The project documentation MUST include a 
       descriptive statement about the scope and 
       duration of support. 
```

I'm bringing this back up one more time because I think it's more clear and direct. But I also won't block on it.

funnelfiasco (approved):
Approved, with a couple of nice-to-have comments.

baseline.yaml
SecurityCRob: maybe instead of "VCS" we sub something like "publicly-available collateral"?

funnelfiasco: That is an acceptably vague compromise to me.

baseline.yaml
```diff
@@ -649,6 +649,27 @@ criteria:
     security_insights_value: # TODO
     scorecard_probe: # TODO
 
+  - id: OSPS-DO-13
+    maturity_level: 1
+    category: Documentation
+    criteria: |
+       The project documentation MUST include a 
+       descriptive statement about the scope and 
+       duration of support.
```
evankanderson, line 658: For some projects, the scope of support is tied (for example) to specific versions.  E.g. Knative has:

https://github.com/knative/community/blob/main/mechanics/RELEASE-SCHEDULE.md#upcoming-releases

> ## Releases supported by community
> 
> | Release | Date       | EOL        | Min K8s Version | Notes |
> |---------| ---------- |------------|-----------------| ----- |
> | 1.16    | 2024-10-22 | 2025-04-22 | 1.29            |       |
> | 1.15    | 2024-07-23 | 2025-01-21 | 1.28            |       |
> 
> ## No longer supported releases
>   
> | Release | Date       | EOL        | Min K8s Version | Notes                                          |
> |---------| ---------- | ---------- |-----------------|------------------------------------------------|
> | 1.14    | 2024-04-23 | 2024-10-22 | 1.28            |                                                |

Having a standard mechanism of communicating this would be great, but it seems like there's a gap here between "having a descriptive statement" and "users can find the descriptive statement".

I realize this is a level 1 expectation, but I'd really like to be able to verify these statements.  Is the idea to use security insights here?

evankanderson: I also don't see a standardized way to report EOL for a particular release branch, which seems like it's what people often actually need (I'm on 3.14 -- do I need to upgrade to 4.x, or will I get a 3.14.1?)

evankanderson: Ah, this looks like it is in #117 .

SecurityCRob: Yeah, support/lifecycle has always been challenging to ID.  Each project handles it slightly differently.  There are a few electronic means I'm aware of (Like OASIS's EoX) that could be explored, but there certainly is no STANDARD for this atm.

SecurityCRob: yes, security insights could be a means to find this info

baseline.yaml
```diff
@@ -649,6 +649,27 @@ criteria:
...
+    objective: |
+       Identify and set expectations for downstream 
+       consumers around what updates the project may
+       or may not provide and when no further updates
+       will be provided.
+    implementation: |
+       Create a status check that checks the project's
+       publicly-available collateral for support 
+       and/or lifecycle statements or machine-readable 
+       metadata.
```
david-a-wheeler, line 668: The "implementation" text is normally how a *project* implements the requirement, not how external parties can *verify* it.

```suggestion
       Provide text identifying which versions/branches of the project
       are being supported and at what level (e.g., improvements welcome,
       bug fixes only, security fixes only, or abandoned).
       Where practical, provide lifecycle statements or machine-readable 
       metadata, e.g., via security-insights.yml project-lifecycle.status.       
```

FYI, here's more info on [security-insights project-lifecycle](https://github.com/ossf/security-insights-spec/blob/main/specification/project-lifecycle.md), but we haven't agreed on how to include links so I haven't suggested it.

david-a-wheeler:
I support the need to document "level of support" in some way, I proposed changes to the implementation text.

I do worry about the word "support". When you're paying for software there's some implied level of support (not always achieved). [Many may bristle at the term "support" at all when they're not being paid to do it](https://www.softwaremaxims.com/blog/not-a-supplier). I can't think of a clearer word though. Suggestions welcome.

funnelfiasco:
> Many may bristle at the term "support" at all when they're not being paid to do it. I can't think of a clearer word though. Suggestions welcome.

"Maintenance" maybe? But I think "support" is well enough understood (and even used) by projects that don't offer paid support agreements that we can use the word. The people who bristle at the word "support" will probably have a similar reaction to this baseline anyway.

baseline.yaml
eddie-knight: Small changes for style. Please double check me here @david-a-wheeler 

```suggestion
	Provide text identifying which iterations of the
	project are being supported, and at what level.
	Some examples include: improvements welcome,
	bug fixes only, security fixes only, or
	abandoned. 

	It is recommended that projects provide lifecycle
	statements or machine-readable  metadata such as
	security-insights.yml project-lifecycle.status.
```

baseline.yaml
evankanderson: I don't think Security Insights has a way to represent this today.

I'm slightly concerned about adding criteria that there's no clear way for projects to fulfil and for us to measure -- once we have a few of those, it seems like it will be hard to avoid adding a bunch more that are not really measurable.

david-a-wheeler:
It might be interesting to look at https://endoflife.date e.g., https://endoflife.date/rails . I don't know anything about this group, but they seem potentially relevant.

baseline.yaml
```diff
@@ -649,6 +649,28 @@ criteria:
     security_insights_value: # TODO
     scorecard_probe: # TODO
 
+  - id: OSPS-DO-13
+    maturity_level: 1
+    category: Documentation
+    criteria: |
```
blkt, line 655: This is breaking CI

```suggestion
    criterion: |
```

puerco:
The proposed changes in this PR will be consolidated in #134

Issue #96, opened 2024-12-02
What happens to deprecated/removed entries?


funnelfiasco:
Mentioned by @eddie-knight in [issue 88](https://github.com/ossf/security-baseline/issues/88#issuecomment-2508663965):

> we should talk about (and document) what the project will do with an ID in the event that an entry is deprecated or removed after an official release. It will be cumbersome to users if an ID is reused for a different entry.

funnelfiasco:
I can draft a PR for this in the next day or two. I don't think it needs a ton of discussion, so we can put together a draft and argue about it.

david-a-wheeler:
In an early stage nothing special, but at some point deprecated/removed entries should have their IDs "reserved" so they're not reused. Otherwise someone who earlier claimed to meet something might suddenly be misunderstood as doing something else.

puerco:
Agree whatever we use to tombstone them it should also signal any automation that the control is gone.

eddie-knight:
Does anyone know how this has been done well elsewhere?

Perhaps we could clear all the keys except for `id` and add a new item signaling the removal?

Issue #86, opened 2024-11-27
Proposed Baseline Maintenance Process


SecurityCRob:
As we start implementing the "1.1" baseline criteria with our pilot projects, we should spend some time thinking about how we will update and maintain the baseline criteria going forward.  I propose the following process (to be added as md file in repo once tex is agreed upon):

1.) Normal text fixes to the criteria will be accepted via PR and reviewed by the baseline project maintainers.  Allowed changes are corrections to spelling/typos, grammar corrections, or enhancements to the supplementary text supporting the criteria including - Objective, Implementation, Control Mappings, and Scorecard/Insights values.  At least two project maintainers must review and approve these changes.
2.) Substantive changes to Criteria including changes to text that alters the originally stated meaning, new Criteria proposals, or removal of Criteria will be documented in GitHub PR(s) and reviewed annually by the Baseline project maintainers.  these changes may reflect changes to global cybersecurity regulations and frameworks or changes in norms around application/project security practices.  Any such substantive changes must be approved by a majority of the project's maintainers.  
3.) Any changes to the Baseline will be reflected within the Compliance Matrix, with new requirements flagged where the Baseline Criteria are appropriate.
4.) Downstream stakeholders will be notified via the project's mailing list on the changes and updates.

mlieberman85:
This seems mostly in line with how we do it in SLSA. I think for number 2 we can make it clear that you're bound to a version of the baseline with the expectation that you align yourself to the latest on a yearly cadence?

Everything else seems straightforward.

funnelfiasco:
This all seems reasonable to me, but for number 2 we might want to have a more frequent cadence for a while. A baseline that's constantly changing isn't much of a baseline, but I expect that real-world adoption will turn up things that we need to address. So maybe a quarterly review cadence (with the understanding that some quarters there will be nothing to review) would be better?

mlieberman85:
Sounds like a good idea. I think this might be just what we have to do for some of these early adopters and then can switch later on once we have some continual monitoring and easy tooling in place or something.

evankanderson:
To what extent do we want to hold defining specific versions of the requirements on tooling to verify those requirements?  I worry about running too far ahead of what we can scale -- i.e. [OSPS-DO-03](https://baseline.openssf.org/#osps-do-03) as writ feels difficult to verify even with modern LLM powers.

mlieberman85:
[...]

PR #116, opened 2024-12-18
Add a maintenance process

funnelfiasco:
This will undoubtedly require a lot of discussion before it's mergeable, but here's a first pass at addressing #86 and #96, which are related issues.

The gist of my initial draft is that instead of publishing the list of criteria directly to baseline.openssf.org, we publish it to a development version page. The index page has information about the project and links to versions. When we cut a new version, we copy the `development.md` to 'v1.2.md` or whatever we call that version.

I took the discussion in #86 as a starting point for the `maintenance.md` page, with one notable change: I described a process that's more continuous, with versions happening at least once a year (but more often if necessary), instead of keeping criteria sitting as open PRs for long periods.

We'll probably want to move some of the contents of `README.md` into the `index.md` file (specifically the description of the criteria) and include a link in README to the website instead, but that can be done later. We probably also want to add a "this is a work-in-progress. don't use it for evaluating your project" to the in-development output, and a version/date tag to the "published" versions.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -1,0 +1,17 @@
+# Open Source Security Baseline
+
+The Open Source Project Security Baseline (OSPS Baseline) is designed to act as a minimum definition of requirements for a project relative to it's maturity level.
+It is maintained by the [OpenSSF Security Baseline SIG](https://github.com/ossf/security-baseline/blob/main/governance/MAINTAINERS.md) according to the [project governance documentation](https://github.com/ossf/security-baseline/blob/main/governance/GOVERNANCE.md).
+
+## Versions
+
+Previous versions are presented for historical reference.
+Downstream consumers of the OSPS Baseline should specify their compliance against a specific version.
+Only the version labeled as "current" should be used for new compliance efforts.
+
+* [In-development version](docs/development)
+* Current version: [v1.0]() released YYYY-MM-DD
+* Previous versions:
+    * [v0.1] released YYYY-MM-DD
+
+Versions are managed according to the [Baseline maintenance process](maintenance).
--- a/docs/index.md
+++ b/docs/index.md
@@ -10,8 +10,10 @@ Downstream consumers of the OSPS Baseline should specify their compliance agains
 Only the version labeled as "current" should be used for new compliance efforts.
 
 * [In-development version](docs/development)
-* Current version: [v1.0]() released YYYY-MM-DD
+<!-- Leave this section out until there are historical and current versions to list.
+* Current version: [2025-04-01]()
 * Previous versions:
-    * [v0.1] released YYYY-MM-DD
+    * (none)
+-->
 
 Versions are managed according to the [Baseline maintenance process](maintenance).
--- a/docs/maintenance.md
+++ b/docs/maintenance.md
@@ -18,6 +18,5 @@ Previous versions of the Baseline will remain available, but are stable and not 
 Retired identifiers will remain in the source yaml files, clearly marked.
 * Substantial changes to the meaning of a criterion will be treated as a new criterion, resulting in a new identifier.
 Minor changes, including a change in level, between Baseline versions will not result in a new identifier.
-* The numeric portion of identifiers are assigned sequentially per category and level.
-Within a category and level, identifiers do not carry additional meaning.
-Moving a criterion between levels will (intentionally) result in assigning a new number.
+* The numeric portion of identifiers are assigned sequentially per category.
+Within a category, identifiers do not carry additional meaning.
--- a/docs/maintenance.md
+++ b/docs/maintenance.md
@@ -1,0 +1,22 @@
+# OSPS Baseline Maintenance Process
+
+* Normal text fixes to the criteria will be accepted via pull request and reviewed by the baseline project maintainers.
+Allowed changes are corrections to spelling/typos, grammar corrections, or enhancements to the supplementary text supporting the criteria, including: Objective, Implementation, Control Mappings, and Scorecard/Insights values.
+At least two project maintainers must review and approve these changes.
+* Substantive changes to Criteria, including changes to text that alters the originally stated meaning, new Criteria proposals, or removal of Criteria will be documented in GitHub PR(s) and reviewed regularly by the Baseline project maintainers for inclusion in the next release.
+These changes may reflect changes to global cybersecurity regulations and frameworks or changes in norms around application/project security practices.
+Any such substantive changes must be approved by a majority of the project's maintainers.
+* As appropriate, but at least annually, the Baseline project maintainers will evaluate the set of criteria and, if necessary, publish a new version of the Baseline.
+Previous versions of the Baseline will remain available, but are stable and not subject to change, except for minor changes to fix technical or typographic errors.
+* Any changes to the Baseline will be reflected within the Compliance Matrix, with new requirements flagged where the Baseline Criteria are appropriate.
+* Versions will follow a calendar-based identification system, using the `YYYY-MM-DD` format.
+* Downstream stakeholders will be notified via the project's mailing list on the changes and updates.
+
+## Identifiers
+
+* Identifiers for retired criteria MUST NOT be reused.
+Retired identifiers will remain in the source yaml files, clearly marked.
+* Substantial changes to the meaning of a criterion will be treated as a new criterion, resulting in a new identifier.
+Minor changes, including a change in level, between Baseline versions will not result in a new identifier.
+* The numeric portion of identifiers are assigned sequentially per category.
+They do not carry additional meaning.
--- a/docs/maintenance.md
+++ b/docs/maintenance.md
@@ -3,7 +3,7 @@
 * Normal text fixes to the criteria will be accepted via pull request and reviewed by the baseline project maintainers.
 Allowed changes are corrections to spelling/typos, grammar corrections, or enhancements to the supplementary text supporting the criteria, including: Objective, Implementation, and Control Mappings.
 At least two project maintainers must review and approve these changes.
-* Substantive changes to Criteria, including changes to text that alters the originally stated meaning, new Criteria proposals, or removal of Criteria will be documented in GitHub PR(s) and reviewed regularly by the Baseline project maintainers for inclusion in the next release.
+* Substantive changes to criteria, including changes to text that alters the originally stated meaning, new criteria proposals, or removal of criteria will be documented in GitHub PR(s) and reviewed regularly by the Baseline project maintainers for inclusion in the next release.
 These changes may reflect changes to global cybersecurity regulations and frameworks or changes in norms around application/project security practices.
 Any such substantive changes must be approved by a majority of the project's maintainers.
 * As appropriate, but at least annually, the Baseline project maintainers will evaluate the set of criteria and, if necessary, publish a new version of the Baseline.
@@ -18,5 +18,6 @@ Previous versions of the Baseline will remain available, but are stable and not 
 Retired identifiers will remain in the source yaml files, clearly marked.
 * Substantial changes to the meaning of a criterion will be treated as a new criterion, resulting in a new identifier.
 Minor changes, including a change in level, between Baseline versions will not result in a new identifier.
-* The numeric portion of identifiers are assigned sequentially per category.
-They do not carry additional meaning.
+* The numeric portion of identifiers are assigned sequentially per category and level.
+Within a category and level, identifiers do not carry additional meaning.
+Moving a criterion between levels will (intentionally) result in assigning a new number.
--- a/docs/maintenance.md
+++ b/docs/maintenance.md
@@ -1,7 +1,7 @@
 # OSPS Baseline Maintenance Process
 
 * Normal text fixes to the criteria will be accepted via pull request and reviewed by the baseline project maintainers.
-Allowed changes are corrections to spelling/typos, grammar corrections, or enhancements to the supplementary text supporting the criteria, including: Objective, Implementation, Control Mappings, and Scorecard/Insights values.
+Allowed changes are corrections to spelling/typos, grammar corrections, or enhancements to the supplementary text supporting the criteria, including: Objective, Implementation, and Control Mappings.
 At least two project maintainers must review and approve these changes.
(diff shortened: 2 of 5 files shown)


funnelfiasco:
IMO we can proceed with this if the general shape is right, even if we need to continue iterating on the specifics of the maintenance process for a bit.

docs/maintenance.md
```diff
@@ -0,0 +1,20 @@
...
+* Any changes to the Baseline will be reflected within the Compliance Matrix, with new requirements flagged where the Baseline Criteria are appropriate.
+* Downstream stakeholders will be notified via the project's mailing list on the changes and updates.
+
+## Identifiers
+
+* Identifiers for retired criteria will not be used.
+* Substantial changes to the meaning of a criterion will be treated as a new criterion, resulting in a new identifier.
+Minor changes, including a change in level, between Baseline versions will not result in a new identifier.
+* The numeric portion of identifiers are assigned sequentially per category.
+They do not carry additional meaning.
```
puerco, line 20: I'm fine having these here for now to keep things moving, but the identifier structure, codes and overall sematics built into the string should be explained in a doc by itself.

docs/maintenance.md
```diff
@@ -0,0 +1,20 @@
...
+At least two project maintainers must review and approve these changes.
+* Substantive changes to Criteria, including changes to text that alters the originally stated meaning, new Criteria proposals, or removal of Criteria will be documented in GitHub PR(s) and reviewed regularly by the Baseline project maintainers.
+These changes may reflect changes to global cybersecurity regulations and frameworks or changes in norms around application/project security practices.
+Any such substantive changes must be approved by a majority of the project's maintainers.
+* As appropriate, but at least annually, the Baseline project maintainers will publish a new version of the Baseline.
+Previous versions of the Baseline will remain avialable, but are not maintained.
+* Any changes to the Baseline will be reflected within the Compliance Matrix, with new requirements flagged where the Baseline Criteria are appropriate.
+* Downstream stakeholders will be notified via the project's mailing list on the changes and updates.
+
+## Identifiers
```
puerco, line 14: Should this be more specific?
```suggestion
## Retiring Identifiers for Deprecated Criteria
```

docs/maintenance.md
```diff
@@ -0,0 +1,20 @@
...
+
+* Identifiers for retired criteria will not be used.
```
puerco, line 16: Let's make explicit the sunsetting process:

```suggestion
* Identifiers for retired criteria MUST be marked in the OSPS baseline markup as deprecated and never reused in future revisions.
```

docs/maintenance.md
evankanderson: I think this applies to all identifiers, not just deprecated ones.  Without a crystal ball, we don't know which identifiers will be deprecated in the future.

docs/maintenance.md
evankanderson, line 16:
```suggestion
* Identifiers for retired criteria MUST NOT be reused.  A list of these retired identifiers will be maintained in (file that contains at least OSPS-AC-06 and OSPS-BR-07)
```

docs/maintenance.md
```diff
@@ -0,0 +1,20 @@
...
+* Substantial changes to the meaning of a criterion will be treated as a new criterion, resulting in a new identifier.
+Minor changes, including a change in level, between Baseline versions will not result in a new identifier.
```
evankanderson, line 18:
```suggestion
Minor changes, including a change in level between Baseline versions will not result in a new identifier.
```

docs/maintenance.md
evankanderson: The point here is, I think, to address [my comment in Slack](https://openssf.slack.com/archives/C07DC6TT2QY/p1734544647124109?thread_ts=1734542444.669299&cid=C07DC6TT2QY)

> So, looking at the series starting with "OSPS-AC-01" through "OSPS-AC-07", AC-01 through AC-04 are all maturity level 1, AC-05 is level 2, and AC-07 is level 3.  If we find a new OSPS-AC that is maturity level 1, what number does it get?

This is probably simply due to historical accident, but it's easy for someone new to come in and see a pattern which is not intentional and cause a mess.

evankanderson (commented):
I'm not sure I quite understand how the branching / releasing strategy will work for future versions -- will we be cutting a release branch for each baseline version?  If so, should `web-publish` be checking out each release branch into a different sub-directory in the future?

(I'm not necessarily asking to do this now, I'm just trying to understand what's being communicated)

docs/maintenance.md
funnelfiasco: The second comma is correct because it sets apart "including a change in level" as a separate clause. Removing that clause, you're left with "Minor changes between Baseline versions will not result in a new identifier."

docs/maintenance.md
funnelfiasco: Evan is correct. This is about how identifiers are governed generally (although it is largely driven by discussions around retirement).

docs/maintenance.md
funnelfiasco: Agreed. That's in the README right now. My intent was to pull it into the index page eventually as we settle on the general shape of this. I'd also be fine putting it in this doc or separate. I think my preference in descending order is:

1. Put the explanation in index.md
2. Put the explanation in maintenance.md
3. Put the explanation in a separate file
4. Leave it in the README for people to find if they go looking for it.

docs/maintenance.md
funnelfiasco: Makes sense. I'll update with some minor revisions to your wording.

funnelfiasco: I like the idea of keeping them in the yaml with a `retired: <BOOLEAN>` entry, which is what I think puerco's comment hinted at.

funnelfiasco:
> I'm not sure I quite understand how the branching / releasing strategy will work for future versions -- will we be cutting a release branch for each baseline version? If so, should `web-publish` be checking out each release branch into a different sub-directory in the future?
> 
> (I'm not necessarily asking to do this now, I'm just trying to understand what's being communicated)

I don't plan on using branching. I can see the appeal, but I think it makes things like publishing the web content more difficult. So the idea is that when we decide we're releasing, we'd build the current state, check in the output as `docs/versions/v<version number>`. The build toolchain would treat it like a static page from there on out.

I didn't include the procedure here because I wanted to keep it to things that consumers care about. What do we think of a `governance/SOP.md` file that has the internal operations for things like this?

docs/maintenance.md
evankanderson: It's about identifier governance in general, because it also covers when and how identifiers can be adjusted between versions (the identifier should "mean" the same thing in the future, so that a tool or document that says `Addresses OSPS-XY-46` without explicitly specifying a baseline version is not ambiguous in the future).

I don't know if it's worth explaining that philosophy in this document or not.  I tend to like to see the "why" behind rules, but some people find it distracting.

docs/maintenance.md
evankanderson: If you want to keep a single file across versions (I think you're gonna regret it, but maybe I'll learn something), you might want `retired: $VERSION` rather than a boolean.

evankanderson: Ah, you'll just check in the file as of a certain point, so then you use text-diff to find the changes over time.  I think that will probably be workable with tools like `blame`.

funnelfiasco: Good point on `retired: $VERSION`. `git blame` would give the same info, but this would be friendlier.

SecurityCRob:
thanks for putting this together @funnelfiasco .  I'll put some thought to it and see if I come up with any new feedback beyond what the crew has given so far.

funnelfiasco:
Repeating from Slack: For the sake of expediency, let's focus on things that are broken or unclear. If something is missing, we can add it in a subsequent PR.

docs/maintenance.md
```diff
@@ -0,0 +1,22 @@
+# OSPS Baseline Maintenance Process
+
+* Normal text fixes to the criteria will be accepted via pull request and reviewed by the baseline project maintainers.
+Allowed changes are corrections to spelling/typos, grammar corrections, or enhancements to the supplementary text supporting the criteria, including: Objective, Implementation, Control Mappings, and Scorecard/Insights values.
+At least two project maintainers must review and approve these changes.
+* Substantive changes to Criteria, including changes to text that alters the originally stated meaning, new Criteria proposals, or removal of Criteria will be documented in GitHub PR(s) and reviewed regularly by the Baseline project maintainers.
```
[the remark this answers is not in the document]
funnelfiasco: Good call, I'll update this slightly.

docs/maintenance.md
```diff
@@ -0,0 +1,22 @@
...
+These changes may reflect changes to global cybersecurity regulations and frameworks or changes in norms around application/project security practices.
+Any such substantive changes must be approved by a majority of the project's maintainers.
+* As appropriate, but at least annually, the Baseline project maintainers will evaluate the set of criteria and, if necessary, publish a new version of the Baseline.
+Previous versions of the Baseline will remain avialable, but are not maintained.
```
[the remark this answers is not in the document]
funnelfiasco: Oh yes, that is _much_ better wording.

funnelfiasco:
Based on feedback, from Evan, I made a few minor wording changes:

````
diff --git a/docs/maintenance.md b/docs/maintenance.md
index ba604b8..6558c93 100644
--- a/docs/maintenance.md
+++ b/docs/maintenance.md
@@ -3,11 +3,11 @@
 * Normal text fixes to the criteria will be accepted via pull request and reviewed by the baseline project maintainers.
 Allowed changes are corrections to spelling/typos, grammar corrections, or enhancements to the supplementary text supporting the criteria, including: Objective, Implementation, Control Mappings, and Scorecard/Insights values.
 At least two project maintainers must review and approve these changes.
-* Substantive changes to Criteria, including changes to text that alters the originally stated meaning, new Criteria proposals, or removal of Criteria will be documented in GitHub PR(s) and reviewed regularly by the Baseline project maintainers.
+* Substantive changes to Criteria, including changes to text that alters the originally stated meaning, new Criteria proposals, or removal of Criteria will be documented in GitHub PR(s) and reviewed regularly by the Baseline project maintainers for inclusion in the next release.
 These changes may reflect changes to global cybersecurity regulations and frameworks or changes in norms around application/project security practices.
 Any such substantive changes must be approved by a majority of the project's maintainers.
 * As appropriate, but at least annually, the Baseline project maintainers will evaluate the set of criteria and, if necessary, publish a new version of the Baseline.
-Previous versions of the Baseline will remain avialable, but are not maintained.
+Previous versions of the Baseline will remain avialable, but are stable and not subject to change, except for minor changes to fix technical or typographic errors.
 * Any changes to the Baseline will be reflected within the Compliance Matrix, with new requirements flagged where the Baseline Criteria are appropriate.
 * Versions will follow a calendar-based identifcation system, using the `YYYY-MM-DD` format.
 * Downstream stakeholders will be notified via the project's mailing list on the changes and updates.
 ````

The only significant change is that I left us the option to fix typos/broken links, etc in released versions.

Also, I noticed a couple of typos that I have just fixed.

**Review by evankanderson — APPROVED** [review]

**Review by eddie-knight — CHANGES_REQUESTED** [review]

File: docs/index.md [hunk 7]
```diff
@@ -0,0 +1,17 @@
...
+Previous versions are presented for historical reference.
+Downstream consumers of the OSPS Baseline should specify their compliance against a specific version.
+Only the version labeled as "current" should be used for new compliance efforts.
+
+* [In-development version](docs/development)
+* Current version: [v1.0]() released YYYY-MM-DD
+* Previous versions:
+    * [v0.1] released YYYY-MM-DD
+
+Versions are managed according to the [Baseline maintenance process](maintenance).
```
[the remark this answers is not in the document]
funnelfiasco: 1. Yes, this is the markdown equivalent of `<a href="maintenance">Baseline maintenance process</a>`. The `maintenance` file will be in the output generated by the docs build process.
2. D'oh. I'll fix that.

docs/maintenance.md
```diff
@@ -0,0 +1,22 @@
...
+* Downstream stakeholders will be notified via the project's mailing list on the changes and updates.
+
+## Identifiers
+
+* Identifiers for retired criteria MUST NOT be reused.
+Retired identifiers will remain in the source yaml files, clearly marked.
+* Substantial changes to the meaning of a criterion will be treated as a new criterion, resulting in a new identifier.
+Minor changes, including a change in level, between Baseline versions will not result in a new identifier.
+* The numeric portion of identifiers are assigned sequentially per category.
+They do not carry additional meaning.
```
eddie-knight, line 22: @SecurityCRob can you suggest a rephrase here to make it fit with your proposed three digit change? (ie, the levels DO carry additional meaning)

funnelfiasco: Are we fully set on the three digit change? The notes from last week say "try doing this with one category, and look at it." (although Crob has done all of the categories) and I, at least, am still not sold on the idea.

In the interests of getting this actually done, let's assume for now that the numbers aren't changing and we can update this if/when they do.

evankanderson: How about this:

```suggestion
* The numeric portion of identifiers are assigned sequentially per category and level.
Within a level (i.e. 201, 202, etc), numbers do not carry a meaning, but should not be re-used across versions.
Moving a criteria between levels will (intentionally) result in assigning a new number, to allow tracking changes to levels over time.
```

docs/maintenance.md
```diff
@@ -0,0 +1,23 @@
...
+
+## Identifiers
+
+* Identifiers for retired criteria MUST NOT be reused.
+Retired identifiers will remain in the source yaml files, clearly marked.
+* Substantial changes to the meaning of a criterion will be treated as a new criterion, resulting in a new identifier.
+Minor changes, including a change in level, between Baseline versions will not result in a new identifier.
+* The numeric portion of identifiers are assigned sequentially per category and level.
+Within a category and level, identifiers do not carry additional meaning.
+Moving a criterion between levels will (intentionally) result in assigning a new number.
```
eddie-knight, line 23:
```suggestion
* The numeric portion of identifiers are assigned sequentially per category.
Within a category, identifiers do not carry additional meaning.
```

Changing per the outcome of #175

eddie-knight (approved):

PR #117, opened 2024-12-18
Suggested wording for end-of-support statement - NEW - OSPS-DO-14

SecurityCRob:
suggested wording for OSPS-DO-14 for End-Of-Support statement

### DIFF
--- a/baseline.yaml
+++ b/baseline.yaml
@@ -652,7 +652,7 @@ criteria:
   - id: OSPS-DO-14
     maturity_level: 3
     category: Documentation
-    criteria: |
+    criterion: |
        The project documentation MUST provide a 
        descriptive statement when releases or 
        versions are no longer supported and that 
--- a/baseline.yaml
+++ b/baseline.yaml
@@ -649,7 +649,7 @@ criteria:
     security_insights_value: # TODO
     scorecard_probe: # TODO
 
-- id: OSPS-DO-14
+  - id: OSPS-DO-14
     maturity_level: 3
     category: Documentation
     criteria: |
--- a/baseline.yaml
+++ b/baseline.yaml
@@ -657,9 +657,6 @@ criteria:
        descriptive statement when releases or 
        versions are no longer supported and that 
        will no longer receive security updates.  
-       
-       This should be provided both in human and 
-       machine-readable formats. 
     objective: |
        To advertise to downstream consumers when software 
        has reached "End of X" [Support, Life, Maintenance, etc.].
--- a/baseline.yaml
+++ b/baseline.yaml
@@ -649,6 +649,30 @@ criteria:
     security_insights_value: # TODO
     scorecard_probe: # TODO
 
+- id: OSPS-DO-14
+    maturity_level: 3
+    category: Documentation
+    criteria: |
+       The project documentation MUST provide a 
+       descriptive statement when releases or 
+       versions are no longer supported and that 
+       will no longer receive security updates.  
+       
+       This should be provided both in human and 
+       machine-readable formats. 
+    objective: |
+       To advertise to downstream consumers when software 
+       has reached "End of X" [Support, Life, Maintenance, etc.].
+    implementation: |
+       Create a status check that checks the project's
+       version control system for support and/or lifecycle
+       statements.  Publishing this in machine-readble 
+       formats is preferable.
+    control_mappings: # TODO
+    security_insights_value: # TODO
+    scorecard_probe: #   
+
+
   - id: OSPS-LE-01
     maturity_level: 2
     category: Legal


baseline.yaml
```diff
@@ -649,6 +649,30 @@ criteria:
...
+    maturity_level: 3
+    category: Documentation
+    criteria: |
+       The project documentation MUST provide a 
+       descriptive statement when releases or 
+       versions are no longer supported and that 
+       will no longer receive security updates.  
+       
+       This should be provided both in human and 
+       machine-readable formats.
```
funnelfiasco, line 662: 1. Do we mean "should" or "SHOULD" in the second sentence? If it's the "normal English 'should'", maybe we actually mean "MUST" instead? (Personally, I lean toward SHOULD, but I want to make sure I understand your intent)
2. As a rule, I don't like the idea of multi-sentence criteria and definitely not multi-paragraph (in the `<p>` sense, not in the English sense). Could it be be re-written as "The project documentation MUST provide human- and machine-readable descriptive statements when releases are no longer supported and will not longer receive security updates"? (As a note, I condensed "releases or versions" to be "releases" because I'm not entirely clear that there's a meaningful distinction between the two). Or, if the intent was for having both human- and machine-readable be a strong suggestion but not a requirement, then we should move the second sentence into the implementation (which it sort of already is)

baseline.yaml
```diff
@@ -649,6 +649,30 @@ criteria:
...
+       This should be provided both in human and 
+       machine-readable formats. 
+    objective: |
+       To advertise to downstream consumers when software 
+       has reached "End of X" [Support, Life, Maintenance, etc.].
+    implementation: |
+       Create a status check that checks the project's
+       version control system for support and/or lifecycle
+       statements.  Publishing this in machine-readble 
+       formats is preferable.
```
david-a-wheeler, line 670: The "implementation" sections should tell projects how to implement it, *not* how to determine if it's true or not.

Maybe we need a new field, similar to the Best Practices badge's "autofill" field, to describe how a criterion could be automatically determined. Note that the Best Practices Badge has a "details" field that explains how *projects* can meet the criterion, and I believe that this is what the current "implementation" field is for.

baseline.yaml
david-a-wheeler, line 662:
```suggestion
```

I like machine-processable formats, but I don't know of one for this use. I certainly don't know of any with widespread acceptance. Security-insights.yml can say when the project ends, but I don't think it can do more.

If there is one, please identify that (at least as an option) in the "implementation" section.

baseline.yaml
evankanderson: I'd _like_ for the implementation guidance to projects to be clear enough that someone defining automation can simply read the implementation description to determine how to check it.  For example, "Publish support information in $FORMAT_X or $FORMAT_Y, with a link at $LOCATION to the support document."

david-a-wheeler: @evankanderson - where we can we should do that. Sometimes, though, it's useful to express how to check for some property in addition to noting how to implement it.

evankanderson: If we have a goal of measuring thousands of projects, I think we'll need this sort of mechanically-readable "contract" between project maintainers (implementation guidance) and baselines / measurement infrastructure. 

Where possible, I'd love to adopt existing patterns for recording this information (even if they are ad hoc), rather than introducing new formats for folks to learn.  (i.e. SECURITY.md heading for vulnerability disclosure) 

Do you have a recommendation about where this information should live?

baseline.yaml
```diff
@@ -649,6 +649,27 @@ criteria:
     security_insights_value: # TODO
     scorecard_probe: # TODO
 
+- id: OSPS-DO-14
```
funnelfiasco, line 652:
```suggestion
  - id: OSPS-DO-14
```

After you rebase, this will fix the CI error.

funnelfiasco (changes requested):
I'm content with the wording, some open questions not withstanding. After the formatting issues are fixed, I think I'll be ready to ✅ .

baseline.yaml
```diff
@@ -649,6 +649,27 @@ criteria:
     security_insights_value: # TODO
     scorecard_probe: # TODO
 
+  - id: OSPS-DO-14
+    maturity_level: 3
+    category: Documentation
+    criteria: |
```
eddie-knight, line 655:
```suggestion
    criterion: |
```

mlieberman85:
I agree with @evankanderson on some of what he posted here. Today there's no real support for end of support/end of life in a programmatic way. There's OpenEoX but as far as I was aware that's still not complete and usable? I could be wrong though.

puerco:
The first openeox spec was published years ago and muy understanding is that its on the process to rehost it in OASIS. Regardless we should favor the machine readable formats as the CRA imposes heavy requirements on long term support and flowing the data from the open source components will help manufacturers understand risk and support obligations better.

baseline.yaml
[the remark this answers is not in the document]
evankanderson: In discussion, it was suggested that a never-to-be-fixed vulnerability is filed when the software reaches end-of-life.

puerco:
The proposed changes in this PR are consolidated in #134

PR #118, opened 2024-12-18
Add dependency selection process - NEW - OSPS-DO-15

SecurityCRob:
added new criteria proposal OSPS-DO-15 - around dependency selection process for projects

### DIFF
--- a/baseline.yaml
+++ b/baseline.yaml
@@ -665,9 +665,15 @@ criteria:
        used in this evaluation helps downstream consumers 
        understand the project's rationale.
     implementation: |
+       Provide documentation on how dependencies are handled.
+       For selection, document what criteria are considered or refer to a document that does so, e.g., the
+       <a href="https://best.openssf.org/Concise-Guide-for-Evaluating-Open-Source-Software">Concise Guide for Evaluating Open Source Software</a>.
+       If a package manager is used to obtain and track dependencies, note that, otherwise
+       explain how dependencies are obtained and tracked.
+    autofill: |
        Create a status check that checks the project's
-       version control system for this dependency 
-       ingestion review.
+       change tracking system (e.g., pull requests or merge requests) for this dependency 
+       ingestion review when a new set of dependencies is proposed.
     control_mappings: # TODO
     security_insights_value: # TODO
     scorecard_probe: #   
--- a/baseline.yaml
+++ b/baseline.yaml
@@ -654,7 +654,7 @@ criteria:
     category: Documentation
     criteria: |
        The project documentation MUST include a 
-       descriptive statement about how the project
+       describe how the project
        selects, obtains, and tracks its dependencies.
     objective: |
        Projects need to understand the security qualities
--- a/baseline.yaml
+++ b/baseline.yaml
@@ -649,6 +649,30 @@ criteria:
     security_insights_value: # TODO
     scorecard_probe: # TODO
 
+      - id: OSPS-DO-15
+    maturity_level: 2
+    category: Documentation
+    criteria: |
+       The project documentation MUST include a 
+       descriptive statement about how the project
+       selects, obtains, and tracks its dependencies.
+    objective: |
+       Projects need to understand the security qualities
+       of the software components they ingest or leverage 
+       as dependencies.  
+       
+       Having a process and policy for what criteria are 
+       used in this evaluation helps downstream consumers 
+       understand the project's rationale.
+    implementation: |
+       Create a status check that checks the project's
+       version control system for this dependency 
+       ingestion review.
+    control_mappings: # TODO
+    security_insights_value: # TODO
+    scorecard_probe: #   
+
+
   - id: OSPS-LE-01
     maturity_level: 2
     category: Legal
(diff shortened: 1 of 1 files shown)


baseline.yaml
```diff
@@ -649,6 +649,30 @@ criteria:
     security_insights_value: # TODO
     scorecard_probe: # TODO
 
+      - id: OSPS-DO-15
```
evankanderson, line 652: Indentation here seems weird.

baseline.yaml
```diff
@@ -649,6 +649,30 @@ criteria:
...
+    maturity_level: 2
+    category: Documentation
+    criteria: |
+       The project documentation MUST include a 
+       descriptive statement about how the project
+       selects, obtains, and tracks its dependencies.
```
evankanderson, line 658: Can we pivot this from "project documentation describes" to something with more teeth?  e.g.

```yaml
- id: OSPS-QA-08
  category: Quality
  criteria: |
    The project has implemented automated review of
    changes to enforce a project-selected dependency
    policy.
```

It's not clear from this criteria whether something that says "if you add a package, remember to update `package.json`, we don't check in `package-lock.json`" could meet the `maturity_level: 2` standard.  (I'd hope not, but it _is_ a descriptive statement about how the projects obtains and tracks its dependencies...)

evankanderson:
Additionally, having a dependency-quality-check as automation would tie in well with [OSPS-QA-04](https://baseline.openssf.org/#osps-qa-04) around requiring status checks to pass.

baseline.yaml
funnelfiasco: These might be two separate things. You can't check against a policy that you don't have, so maybe having _any_ policy is level 1 and your suggestion is level 2?

baseline.yaml
SecurityCRob: i should have that fixed now

baseline.yaml
```diff
@@ -649,6 +649,30 @@ criteria:
     security_insights_value: # TODO
     scorecard_probe: # TODO
 
+  - id: OSPS-DO-15
+    maturity_level: 2
+    category: Documentation
+    criteria: |
+       The project documentation MUST include a 
+       descriptive statement about how the project
```
david-a-wheeler, line 657:
```suggestion
       describe how the project
```

A "statement" often implies something short, that is, a sentence or "single declaration or remark". We should not limit them to only give one sentence, and this change makes the criterion text shorter too.

baseline.yaml
```diff
@@ -649,6 +649,30 @@ criteria:
...
+       of the software components they ingest or leverage 
+       as dependencies.  
+       
+       Having a process and policy for what criteria are 
+       used in this evaluation helps downstream consumers 
+       understand the project's rationale.
+    implementation: |
+       Create a status check that checks the project's
+       version control system for this dependency 
+       ingestion review.
```
david-a-wheeler, line 670:
```suggestion
       Provide documentation on how dependencies are handled.
       For selection, document what criteria are considered or refer to a document that does so, e.g., the
       <a href="https://best.openssf.org/Concise-Guide-for-Evaluating-Open-Source-Software">Concise Guide for Evaluating Open Source Software</a>.
       If a package manager is used to obtain and track dependencies, note that, otherwise
       explain how dependencies are obtained and tracked.
    autofill:
       Create a status check that checks the project's
       change tracking system (e.g., pull requests or merge requests) for this dependency 
       ingestion review when a new set of dependencies is proposed.
```

The term "version control system" (VCS) doesn't seem correct here. E.g., The tool "git" is the most popular VCS, but it won't have this data. GitHub and GitLab typically do, but they're more than git and provide more than VCS.

funnelfiasco (approved):

baseline.yaml
```diff
@@ -649,6 +649,36 @@ criteria:
     security_insights_value: # TODO
     scorecard_probe: # TODO
 
+  - id: OSPS-DO-15
+    maturity_level: 2
+    category: Documentation
+    criteria: |
+       The project documentation MUST include a 
+       describe how the project
```
eddie-knight, line 657:
```suggestion
       description of how the project
```

david-a-wheeler (approved):

baseline.yaml
```diff
@@ -649,6 +649,36 @@ criteria:
     security_insights_value: # TODO
     scorecard_probe: # TODO
 
+  - id: OSPS-DO-15
+    maturity_level: 2
+    category: Documentation
+    criteria: |
```
blkt, line 655: **nit:** This should be breaking CI, I'm confused why it doesn't 🤔

mlieberman85: What do you mean?

funnelfiasco: @mlieberman85 we recently switch the yaml to use `criterion` instead of `criteria`. I believe that was merged after the last commit to this PR, which is why the CI hasn't failed as @blkt expects, but it should fail on the next commit if it's not fixed.

eddie-knight: I gave this a re-run, no break... The unknown field just gets silently ignored.

We should fix it though.

```suggestion
    criterion: |
```

puerco:
The changes proposed by this PR are consolidated in #134

Issue #112, opened 2024-12-16
Create a style guide


funnelfiasco:
Since we have many people working on this, and the style is relatively important for this sort of work, we should make our style explicit. Right now, we have an implicit style.

SecurityCRob:
the group was SUPER excited to have a style guide someday (soon) STRONG +1's all around!

eddie-knight:
Ey ey ey, I hear you

funnelfiasco:
Submitted a funding proposal for a technical writer (ossf/tac#511)

eddie-knight:
Resolved by #533

PR #119, opened 2024-12-18
Project roles and responsibilities - NEW - OSPS-DO-16

SecurityCRob:
added suggestion for criteria OSPS-DO-16 which covers project roles & responsibilities

### DIFF
--- a/baseline.yaml
+++ b/baseline.yaml
@@ -665,8 +665,16 @@ criteria:
         review and demonstrates maturity of project
         processes.
     implementation: |
+       Document the various roles and responsibilities in your project.
+       Many OSS projects have "Maintainers" who have the authority
+       to directly accept changes, and contrast this with "Contributors"
+       who do not. If a project grows large there may be many
+       different roles.
+    autofill: |
        Create a status check that checks the project's
        version control system for this documentation.
+       Search for terms like "roles and responsibilities",
+       "duties", and "expectations" (if in English).
     control_mappings: # TODO
     security_insights_value: # TODO
     scorecard_probe: #   
--- a/baseline.yaml
+++ b/baseline.yaml
@@ -658,7 +658,7 @@ criteria:
     objective: |
        Projects need to document the Roles and 
        Responsibilities of the project to provide for 
-       Seperation of Duties, Dual Control, and other 
+       Separation of Duties, Dual Control, and other 
        requirements.
 
        Providing this documentation allows external
--- a/baseline.yaml
+++ b/baseline.yaml
@@ -657,7 +657,7 @@ criteria:
        and Responsibilities for members of the project.
     objective: |
        Projects need to document the Roles and 
-       Responsibilities of the project to provide for 
+       Responsibilities of the project's members to provide for 
        Separation of Duties, Dual Control, and other 
        requirements.
 
--- a/baseline.yaml
+++ b/baseline.yaml
@@ -649,6 +649,29 @@ criteria:
     security_insights_value: # TODO
     scorecard_probe: # TODO
 
+  - id: OSPS-DO-16  
+    maturity_level: 2
+    category: Documentation
+    criteria: |
+       The project documentation MUST include the Roles
+       and Responsibilities for members of the project.
+    objective: |
+       Projects need to document the Roles and 
+       Responsibilities of the project to provide for 
+       Seperation of Duties, Dual Control, and other 
+       requirements.
+
+       Providing this documentation allows external
+        review and demonstrates maturity of project
+        processes.
+    implementation: |
+       Create a status check that checks the project's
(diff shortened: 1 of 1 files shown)


SecurityCRob:
Not 100% convinced this is a "docs", but open to alternate landing suggestions

evankanderson:
It feels like there is probably a "governance" section waiting to happen.  I'd nominate OSPS-DO-01, OSPS-DO-02, OSPS-DO-06, OSPS-DO-11 for inclusion, or it could be combined with the LE- line if desired.

baseline.yaml
```diff
@@ -649,6 +649,29 @@ criteria:
...
+    maturity_level: 2
+    category: Documentation
+    criteria: |
+       The project documentation MUST include the Roles
+       and Responsibilities for members of the project.
+    objective: |
+       Projects need to document the Roles and 
+       Responsibilities of the project to provide for 
+       Seperation of Duties, Dual Control, and other 
+       requirements.
```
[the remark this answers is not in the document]
SecurityCRob: do-11 is about access review of those that get the commit-bit, this is broader, "document whom can do what" within the project

baseline.yaml
```diff
@@ -649,6 +649,29 @@ criteria:
...
 
+  - id: OSPS-DO-16  
+    maturity_level: 2
+    category: Documentation
+    criteria: |
+       The project documentation MUST include the Roles
+       and Responsibilities for members of the project.
+    objective: |
+       Projects need to document the Roles and 
+       Responsibilities of the project to provide for
```
funnelfiasco, line 660: Should "of the project" be "of project members" or "of project participants" or something like that?

funnelfiasco:
I'm in favor of creating a Governance category, as proposed.

funnelfiasco:
A general comment on this PR: there are a lot of capital letters in places I don't expect. Maybe that's something for Eddie to include in the style guide (#112)

baseline.yaml
SecurityCRob: good suggestion!  I'll change it.

baseline.yaml
```diff
@@ -649,6 +649,29 @@ criteria:
...
+       Responsibilities of the project's members to provide for 
+       Separation of Duties, Dual Control, and other 
+       requirements.
+
+       Providing this documentation allows external
+        review and demonstrates maturity of project
+        processes.
+    implementation: |
+       Create a status check that checks the project's
+       version control system for this documentation.
```
david-a-wheeler, line 669:
```suggestion
    implementation: |
       Document the various roles and responsibilities in your project.
       Many OSS projects have "Maintainers" who have the authority
       to directly accept changes, and contrast this with "Contributors"
       who do not. If a project grows large there may be many
       different roles.
    autofill: |
       Create a status check that checks the project's
       version control system for this documentation.
       Search for terms like "roles and responsibilities",
       "duties", and "expectations" (if in English).
```

david-a-wheeler:
What about one-person projects? It's obviously possible to document roles & responsibilities even in that case, in the hopes that the project will grow, but is that important while there's only one person?

funnelfiasco:
> What about one-person projects? It's obviously possible to document roles & responsibilities even in that case, in the hopes that the project will grow, but is that important while there's only one person?

One-person projects are unlikely to care about the baseline anyway. If nothing else, they wouldn't be able to meet the pending OSPS-AC-08 (#123) requirement.

evankanderson:
> > What about one-person projects? It's obviously possible to document roles & responsibilities even in that case, in the hopes that the project will grow, but is that important while there's only one person?
> 
> One-person projects are unlikely to care about the baseline anyway. If nothing else, they wouldn't be able to meet the pending OSPS-AC-08 (#123) requirement.

Yeah -- I'm wondering whether we should have a profile for single-person projects that excludes certain types of practices that only make sense for teams (like lowest privileges).

At the same time, it seems fair to cap a 1-person project to e.g. maturity level 1, because there's a single point of failure for the whole project if that person has life happen.

david-a-wheeler:
I'm +1 on creating a "governance" section. Seems reasonable.

baseline.yaml
```diff
@@ -649,6 +649,37 @@ criteria:
     security_insights_value: # TODO
     scorecard_probe: # TODO
 
+  - id: OSPS-DO-16  
+    maturity_level: 2
+    category: Documentation
+    criteria: |
```
blkt, line 655: **nit:** This should be breaking CI, I'm confused why it doesn't 🤔

funnelfiasco: I don't think the CI has been re-run since that change was made. If it were to run now, the CI check should fail.

puerco:
The changes in this PR are consolidated in #134

PR #120, opened 2024-12-18
Project risk assessment process/docs - NEW - OSPS-DO-17

SecurityCRob:
added proposal for OSPS-DO-18 for project's risk assessment process & docs

### DIFF
--- a/baseline.yaml
+++ b/baseline.yaml
@@ -650,7 +650,7 @@ criteria:
     scorecard_probe: # TODO
 
   - id: OSPS-DO-17   
-    maturity_level: 1
+    maturity_level: 2
     category: Documentation
     criteria: |
        The project MUST perform a risk assessment to 
--- a/baseline.yaml
+++ b/baseline.yaml
@@ -653,8 +653,8 @@ criteria:
     maturity_level: 2
     category: Documentation
     criteria: |
-       The project MUST perform a risk assessment to 
-       understand the most likely and impactful problems 
+       The project MUST perform a security risk assessment to 
+       understand the most likely and impactful potential security problems 
        that could occur within the software.
     objective: |
        Projects need to document know threats and 
--- a/baseline.yaml
+++ b/baseline.yaml
@@ -661,9 +661,15 @@ criteria:
        evaluate the likehood and impact of the risks the 
        project identifies as being of the most concern.
     implementation: |
+       Identify a set of security threats, then estimate their
+       likelihood and impact.
+    autofill: |
        Create a status check that checks the project's
        version control system for this risk assessment, 
        including plans to address any identified risks.
+       Search for terms like "risk assessment" or
+       "risk management plan" where terms like
+       "security" or "vulnerability" are nearby.
     control_mappings: # TODO
     security_insights_value: # TODO
     scorecard_probe: #   
--- a/baseline.yaml
+++ b/baseline.yaml
@@ -649,6 +649,26 @@ criteria:
     security_insights_value: # TODO
     scorecard_probe: # TODO
 
+  - id: OSPS-DO-17   
+    maturity_level: 1
+    category: Documentation
+    criteria: |
+       The project MUST perform a risk assessment to 
+       understand the most likely and impactful problems 
+       that could occur within the software.
+    objective: |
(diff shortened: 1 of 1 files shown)


baseline.yaml
```diff
@@ -649,6 +649,26 @@ criteria:
     security_insights_value: # TODO
     scorecard_probe: # TODO
 
+  - id: OSPS-DO-17   
+    maturity_level: 1
```
evankanderson, line 653: If this is a maturity level 1 requirement, it feels like we should be more descriptive of how high this wall is.  For example, is having 2 paragraphs under a heading in `SECURITY.md` sufficient, or is a stand-alone document required?  Are there specific frameworks that should be referenced, or threat models documented?

funnelfiasco: I agree. This needs to be explained in much more detail, particularly if it's level 1. Depending on what we mean by "risk assessment", I'm inclined to think this is maybe a level 2 requirement?

funnelfiasco (changes requested):

baseline.yaml
TheFoxAtWork: concur with level 2 - but we should double check against what other criteria would be considered part of a risk assessment by the project (threat model, security analysis, etc.) to ensure we're not setting this too low if components that contribute are at level 3

baseline.yaml
```diff
@@ -649,6 +649,26 @@ criteria:
     security_insights_value: # TODO
     scorecard_probe: # TODO
 
+  - id: OSPS-DO-17   
+    maturity_level: 2
+    category: Documentation
+    criteria: |
+       The project MUST perform a risk assessment to 
+       understand the most likely and impactful problems 
+       that could occur within the software.
```
david-a-wheeler, line 658:
```suggestion
       The project MUST perform a security risk assessment to 
       understand the most likely and impactful potential security problems 
       that could occur within the software.
```

We must clarify we're talking about security.

To some managers, a "risk assessment" for a project is determining the risks to failing to meet schedule and/or budget. Functionality and security might be gleefully traded away for schedule and budget. I don't think that's what you mean. Also say "potential... problem"; a risk is a potential problem, and if it's an actual problem it's no longer just a risk.

baseline.yaml
```diff
@@ -649,6 +649,26 @@ criteria:
...
+    objective: |
+       Projects need to document know threats and 
+       evaluate the likehood and impact of the risks the 
+       project identifies as being of the most concern.
+    implementation: |
+       Create a status check that checks the project's
+       version control system for this risk assessment, 
+       including plans to address any identified risks.
```
david-a-wheeler, line 666:
```suggestion
    implementation: |
       Identify a set of security threats, then estimate their
       likelihood and impact.
    autofill: |
       Create a status check that checks the project's
       version control system for this risk assessment, 
       including plans to address any identified risks.
       Search for terms like "risk assessment" or
       "risk management plan" where terms like
       "security" or "vulnerability" are nearby.
```

david-a-wheeler:
The term "threat" is really vague, it's not clear what level we're discussing. E.g., is this a list of threats?: "Unintentional vulnerability in our code", "Malicious vulnerability in our code", "Vulnerability in dependency", "Insecure design"? Can we suggest a specific list as a starting point?

Can we identify some OSS projects as examples who have done this? For all our criteria we should be pointing to projects that have done it.

SecurityCRob:
We will create a new category for "Security Assessments" and move this there.  This will be OSPS-SA-01.

baseline.yaml
```diff
@@ -649,6 +649,32 @@ criteria:
...
 
+  - id: OSPS-DO-17   
+    maturity_level: 2
+    category: Documentation
+    criteria: |
+       The project MUST perform a security risk assessment to 
+       understand the most likely and impactful potential security problems 
+       that could occur within the software.
+    objective: |
+       Projects need to document know threats and
```
david-a-wheeler, line 660:
```suggestion
       Projects need to document threats and 
```

david-a-wheeler: It's not clear how you document unknown threats :-). Let's omit unnecessary words.

puerco:

PR #121, opened 2024-12-18
Add proposal for threat modeling/attack analysis - NEW - OSPS-DO-18

SecurityCRob:
added proposal for Threat modeling, attack surface analysis, and/or data-flow analysis as part of process & docs

### DIFF
--- a/baseline.yaml
+++ b/baseline.yaml
@@ -667,6 +667,11 @@ criteria:
         reducing potential attack surface and to harden 
         the software from specific attacks.
     implementation: |
+       Select a threat modeling approach such as STRIDE, DREAD, PASTA, or VAST, then apply it.
+       This will typically involve identifying the scope and purpose of the system,
+       identifying its assets (which need protection), examining the architecture for threats,
+       determining their likelihood and impact, and selecting mitigation strategies.
+     autofill: |
        Create a status check that checks the project's
        version control system for documented threat 
        modeling, attack surface analysis, and data flow analysis.
--- a/baseline.yaml
+++ b/baseline.yaml
@@ -649,7 +649,7 @@ criteria:
     security_insights_value: # TODO
     scorecard_probe: # TODO
 
-      - id: OSPS-DO-18   
+  - id: OSPS-DO-18   
     maturity_level: 2
     category: Documentation
     criteria: |
--- a/baseline.yaml
+++ b/baseline.yaml
@@ -649,6 +649,33 @@ criteria:
     security_insights_value: # TODO
     scorecard_probe: # TODO
 
+      - id: OSPS-DO-18   
+    maturity_level: 2
+    category: Documentation
+    criteria: |
+       The project MUST perform threat modeling and
+       attack surface analysis to understand and protect
+       against critical code paths, functions, and interactions 
+       with the system.
+    objective: |
+       Projects need to conduct threat modeling, attack 
+       surface analysis, and data-flow analysis in order
+       to understand, document, and plan protections
+       to avoid future explotation of threats and weaknesses.
+
+       Identifying these areas helps the project plan on
+        reducing potential attack surface and to harden 
+        the software from specific attacks.
+    implementation: |
+       Create a status check that checks the project's
+       version control system for documented threat 
+       modeling, attack surface analysis, and data flow analysis.
+    control_mappings: # TODO
+    security_insights_value: # TODO
+    scorecard_probe: #   
+
+ 
+
   - id: OSPS-LE-01
     maturity_level: 2
     category: Legal
--- a/baseline.yaml
+++ b/baseline.yaml
@@ -655,8 +655,8 @@ criteria:
     criteria: |
        The project MUST perform threat modeling and
        attack surface analysis to understand and protect
-       against critical code paths, functions, and interactions 
-       with the system.
+       against attacks on critical code paths, functions, and interactions 
+       within the system.
     objective: |
        Projects need to conduct threat modeling, attack 
        surface analysis, and data-flow analysis in order
--- a/baseline.yaml
+++ b/baseline.yaml
@@ -670,6 +670,11 @@ criteria:
        Create a status check that checks the project's
        version control system for documented threat 
        modeling, attack surface analysis, and data flow analysis.
+       
+       The location of the written threat model MAY be expressed using [the `security-artifacts.threat-model`
+       fields in `SECURITY-INSIGHTS.yaml`](https://github.com/ossf/security-insights-spec/blob/main/specification/security-artifacts.md),
+       or via a [SPDX external reference](https://spdx.github.io/spdx-spec/v3.0.1/model/Core/Vocabularies/ExternalRefType/)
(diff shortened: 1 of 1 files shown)


baseline.yaml
```diff
@@ -649,6 +649,33 @@ criteria:
     security_insights_value: # TODO
     scorecard_probe: # TODO
 
+      - id: OSPS-DO-18
```
puerco, line 652: Watch the indent here

```suggestion
  - id: OSPS-DO-18   
```

baseline.yaml
```diff
@@ -649,6 +649,33 @@ criteria:
...
+       to understand, document, and plan protections
+       to avoid future explotation of threats and weaknesses.
+
+       Identifying these areas helps the project plan on
+        reducing potential attack surface and to harden 
+        the software from specific attacks.
+    implementation: |
+       Create a status check that checks the project's
+       version control system for documented threat 
+       modeling, attack surface analysis, and data flow analysis.
```
puerco, line 672: The status check is the job tools that want to automate the compliance check, the implementation should guide the project to comply with the criteria, I think should be summarized as:

1. Write the threat model
2. Publish it somewhere
3. Leave behind some pointers linking the repository to it, form example in security insights or using an [SPDX external reference](https://spdx.github.io/spdx-spec/v3.0.1/model/Core/Vocabularies/ExternalRefType/) of type `securityThreatModel`.

evankanderson:
```suggestion
       Create a status check that checks the project's
       version control system for documented threat 
       modeling, attack surface analysis, and data flow analysis.
       
       The location of the written threat model MAY be expressed using [the `security-artifacts.threat-model`
       fields in `SECURITY-INSIGHTS.yaml`](https://github.com/ossf/security-insights-spec/blob/main/specification/security-artifacts.md),
       or via a [SPDX external reference](https://spdx.github.io/spdx-spec/v3.0.1/model/Core/Vocabularies/ExternalRefType/)
       of type `securityThreatModel`, for example.
```

evankanderson (commented):
+1 on providing specific examples of how to locate the threat model.  I think it's fine to grow more examples over time, but leaving this a blank slate makes it hard for tools and project owners to converge on a small set of solutions rather than balls of markdown.

baseline.yaml
evankanderson: The `artifacts.threat-model` field has 3 sub-fields:

* `threat-model-created` (bool)
* `evidence-url` (array of IRI / RFC 3987)
* `comment` (<560 chars)

I'd expect many `evidence-url`s to be outside of VCS, but the idea is to put the *pointer* into VCS, so we can find and fetch the external resource.

baseline.yaml
```diff
@@ -649,6 +649,38 @@ criteria:
...
     scorecard_probe: # TODO
 
+  - id: OSPS-DO-18   
+    maturity_level: 2
+    category: Documentation
+    criteria: |
+       The project MUST perform threat modeling and
+       attack surface analysis to understand and protect
+       against critical code paths, functions, and interactions 
+       with the system.
```
david-a-wheeler, line 659:
```suggestion
       against attacks on critical code paths, functions, and interactions 
       within the system.
```
Presumably we don't want to prevent the use of critical code paths :-).

baseline.yaml
```diff
@@ -649,6 +649,38 @@ criteria:
...
+    objective: |
+       Projects need to conduct threat modeling, attack 
+       surface analysis, and data-flow analysis in order
```
david-a-wheeler, line 662:
```suggestion
       Projects need to conduct threat modeling and attack 
       surface analysis in order
```

Just say threat modeling. Data flow analysis typically follows on as part of threat modeling, but you can do lots of data flow analysis that has nothing to do with security, and there's no need to require it. Just require the threat modeling and I think you're covered.

baseline.yaml
```diff
@@ -649,6 +649,38 @@ criteria:
...
+       to understand, document, and plan protections
+       to avoid future explotation of threats and weaknesses.
+
+       Identifying these areas helps the project plan on
+        reducing potential attack surface and to harden 
+        the software from specific attacks.
+    implementation: |
+       Create a status check that checks the project's
+       version control system for documented threat 
+       modeling, attack surface analysis, and data flow analysis.
```
david-a-wheeler, line 672:
```suggestion
    implementation: |
       Select a threat modeling approach such as STRIDE, DREAD, PASTA, or VAST, then apply it.
       This will typically involve identifying the scope and purpose of the system,
       identifying its assets (which need protection), examining the architecture for threats,
       determining their likelihood and impact, and selecting mitigation strategies.
     autofill: |
       Create a status check that checks the project's
       version control system for documented threat 
       modeling, attack surface analysis, and data flow analysis.
```

david-a-wheeler (approved):
I'm thinking this could be clarified further, but having *specific* examples is a step forward.

I'm wavering on whether or not this should be level 2 or level 3. I can see an argument for level 3. However, a simple basic threat modeling analysis isn't *that* hard, especially since we don't specify the level of depth. So I'm just saying "yes" here.

eddie-knight:
As much as I'm usually a proponent of threat models, I'm not a fan of the complexity here. Definitely not for level 2.

We don't want to hold small project teams to an overly rigorous standard, and even the simplest threat model processes can become _very_ cumbersome. In contrast, the self-assessments we've discussed previously could accomplish most of our goals without the added rigor of a threat model. 

The process we've been following already seems to cover 2/3 of the goals here:

- [DO-03](https://baseline.openssf.org/#osps-do-03) covers the _"scope and purpose of the system"_ (though perhaps this could be more direct)
- [DO-07](https://baseline.openssf.org/#osps-do-07) and [DO-09](https://baseline.openssf.org/#osps-do-09) cover actors, actions, inputs, and outputs— similar to _"identifying its assets (which need protection), examining the architecture for threats"_
- A gap still exists on documenting _"threats, determining their likelihood and impact, and selecting mitigation strategies."_

mlieberman85:
I agree with Eddie. I still think we have something like a risk profile missing here. For example I can see us requiring this for Level 2 if Low Risk projects are excluded until Level 3 or something like that.

SecurityCRob:
We will create a new category for "Security Assessments" and move this there.  This will be OSPS-SA-02.

baseline.yaml
```diff
@@ -649,6 +649,43 @@ criteria:
...
+
+       Identifying these areas helps the project plan on
+        reducing potential attack surface and to harden 
+        the software from specific attacks.
+    implementation: |
+       Select a threat modeling approach such as STRIDE, DREAD, PASTA, or VAST, then apply it.
+       This will typically involve identifying the scope and purpose of the system,
+       identifying its assets (which need protection), examining the architecture for threats,
+       determining their likelihood and impact, and selecting mitigation strategies.
+     autofill: |
```
evankanderson, line 674:
```suggestion
    autofill: |
```

evankanderson: I'm not sure that `autofill` is in our schema.

puerco:

PR #124, opened 2024-12-19
Add contributor guidance (OSPS-DO-02)

evankanderson:
Per [discussion on slack](https://openssf.slack.com/archives/C07DC6TT2QY/p1734631367135579?thread_ts=1734621386.086959&cid=C07DC6TT2QY)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,26 @@
+# Contributing to the Security Baseline
+
+## Contributing to the Criteria:
+
+Currently, all the baseline criteria are in `baseline.yaml`; follow the structure
+[in the README](./README.md#baseline-structure) when proposing new entries.
+
+## Contributing to the Tooling:
+
+The baseline is published to https://baseline.openssf.org/ (via GitHub Pages) via
+Jekyll (a static site generator) using scripts from `./cmd` and formatting from
+`./docs`, using GitHub Actions.
+
+## PR guidelines
+
+All changes to the repository should be made via PR
+([OSPS-AC-03](https://baseline.openssf.org/#osps-ac-03)).  In addition to a clear
+title and descriptive commit message, PRs MUST meet the following criteria:
+
+* DCO signoff (via `git commit -s` -- [OSPS-LE-01](https://baseline.openssf.org/#osps-le-01))
+* All checks must pass ([OSPS-QA-04](https://baseline.openssf.org/#osps-qa-04))
+
+## Maintainer Status
+
+See [./governance/GOVERNANCE.md](./governance/GOVERNANCE.md#maintainer-status) for
+the process of achieving maintainer status on the project.
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ Each entry has the following values:
 
 ## Contribution, Governance, & Security
 
-Contributions are always welcome via pull request or GitHub Discussions. Refer to the governance documentation for information about [how the project operates] and [how to report security-related issues].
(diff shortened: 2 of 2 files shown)


SecurityCRob (approved):

funnelfiasco (approved):
~~LGTM~~ (edit because I noticed the DCO after)

funnelfiasco (changes requested):
@evankanderson you need to signoff on the commit before we can merge it.

~~(And so we should probably add that to this file as well)~~
Nevermind the above. I see that you added it, you just didn't follow it. 😄

evankanderson:
> @evankanderson you need to signoff on the commit before we can merge it.
> 
> ~(And so we should probably add that to this file as well)~ Nevermind the above. I see that you added it, you just didn't follow it. 😄

... I was just checking, of course!  😉 

Fixed.

CONTRIBUTING.md
```diff
@@ -0,0 +1,26 @@
...
+`./docs`, using GitHub Actions.
+
+## PR guidelines
+
+All changes to the repository should be made via PR
+([OSPS-AC-03](https://baseline.openssf.org/#osps-ac-03)).  In addition to a clear
+title and descriptive commit message, PRs MUST meet the following criteria:
+
+* DCO signoff (via `git commit -s` -- [OSPS-LE-01](https://baseline.openssf.org/#osps-le-01))
+* All checks must pass ([OSPS-QA-04](https://baseline.openssf.org/#osps-qa-04))
```
funnelfiasco, line 21: We might want to leave the links out of this, since #113 will break them anyway. I think we're fine without the citations, but if you feel strongly about leaving them in, we should decide if we want to have them be hyperlinks or just bare text references.

CONTRIBUTING.md
```diff
@@ -0,0 +1,26 @@
...
+## Contributing to the Tooling:
+
+The baseline is published to https://baseline.openssf.org/ (via GitHub Pages) via
+Jekyll (a static site generator) using scripts from `./cmd` and formatting from
+`./docs`, using GitHub Actions.
+
+## PR guidelines
+
+All changes to the repository should be made via PR
+([OSPS-AC-03](https://baseline.openssf.org/#osps-ac-03)).  In addition to a clear
```
david-a-wheeler, line 17:
```suggestion
All changes to the repository should be made via a pull request (PR)
(this enables us to meet our own requirement [OSPS-AC-03](https://baseline.openssf.org/#osps-ac-03)).
In addition to a clear
```

CONTRIBUTING.md
evankanderson: As "customer zero", we should figure out what the best practices are here.  Linking to the criteria seems helpful in providing background information, but only if the URLs are stable enough that they don't just turn into dead links.

(And, I think that's something worth thinking about as we proceed...)

evankanderson: #113 did not break the links, BTW, because the anchor text for the criteria stayed the same.

evankanderson: (Also, bare text references makes it hard to find `baseline.openssf.org` from the github repo, because we don't have the homepage metadata filled in...)

evankanderson:
I was re-reading the README and saw that we were recommending GitHub Discussions as a way to reach the team (but we don't have discussions turned on).  I've changed that to filing GitHub issues, and also provided a link to the OpenSSF slack channel, since that wasn't immediately discoverable from the repository.

funnelfiasco (approved):
LGTM, thanks!

PR #126, opened 2025-01-03
Added new category for threat modeling

eddie-knight:
Adding category "Security Assessments" to match the phrasing in Security Insights v2.

This will need to be reorganized in the yaml (lots of that still to do)

This will allow us to extend the category to cover all topics necessary to incrementally produce a complete threat model based on what is reasonable for each maturity level.

Inspired by, but in lieu of, #121

### DIFF
--- a/baseline.yaml
+++ b/baseline.yaml
@@ -507,7 +507,7 @@ criteria:
     control_mappings: # TODO
     security_insights_value: # TODO
 
-  - id: OSPS-TM-07
+  - id: OSPS-TM-02
     maturity_level: 2
     category: Threat Modeling
     criterion: |
@@ -552,7 +552,7 @@ criteria:
     control_mappings: # TODO
     security_insights_value: # TODO
 
-  - id: OSPS-TM-09
+  - id: OSPS-TM-03
     maturity_level: 3
     category: Threat Modeling
     criterion: |
@@ -961,6 +961,32 @@ criteria:
     control_mappings: # TODO
     security_insights_value: # TODO
     scorecard_probe: # sastToolRunsOnAllCommits 
+
+  - id: OSPS-TM-04
+    maturity_level: 2
+    category: Threat Modeling
+    criterion: |
+      The project documentation MUST include a
+      document that details the assets, threats,
+      and mitigations for the project.
+    rationale: |
+      Provide a structured approach to identifying
+      and addressing security threats to the
+      project, enabling contributors and users to
+      understand the risks and mitigations in place.
+    details: |
+      Create a threat model or assessment for the 
+      project. This will identify the assets,
+      threats, and mitigations for the project.
+      Include information on the potential threats
+      to the project's assets and the measures in
+      place to mitigate those threats.
+
+      The authoring process will be simplified if
+      including or referencing the output from
+      OSPS-TM-01, OSPS-TM-02, and OSPS-TM-03.
+    control_mappings: # TODO
+    security_insights_value: # TODO
 
 # # # #
 #
--- a/baseline.yaml
+++ b/baseline.yaml
@@ -404,16 +404,17 @@ criteria:
     scorecard_probe:
       - # None, may not be suitable
 
-  - id: OSPS-DO-03
+  - id: OSPS-TM-01
     maturity_level: 2
-    category: Documentation
+    category: Threat Modeling
     criterion: |
       The project documentation MUST provide user
       guides for all basic functionality.
     rationale: |
-      Ensure that users have a clear and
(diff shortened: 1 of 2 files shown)


SecurityCRob:
Do you feel this new category would replace the proposed  https://github.com/ossf/security-baseline/pull/120 and https://github.com/ossf/security-baseline/pull/121 ?

eddie-knight:
@SecurityCRob Yeah, there may be more to consider for the category, but I think this moves it in the right direction.

baseline.yaml
```diff
@@ -404,16 +404,17 @@ criteria:
     scorecard_probe:
       - # None, may not be suitable
 
-  - id: OSPS-DO-03
+  - id: OSPS-TM-01
```
eddie-knight, line 407:
```suggestion
  - id: OSPS-SA-01
```

baseline.yaml
```diff
@@ -550,19 +552,19 @@ criteria:
     control_mappings: # TODO
     security_insights_value: # TODO
 
-  - id: OSPS-DO-09
+  - id: OSPS-TM-03
```
eddie-knight, line 555:
```suggestion
  - id: OSPS-SA-03
```

funnelfiasco:
I agree that "Security Assessments" is a better term than "Threat Modeling" here, but the concept seems wrong to me as applied in the current PRs. Everything in the baseline is useful for downstream performing a security assessment, so why are these things that are clearly documentation special? I would expect the "Security Assessments" category to be about the security assessments performed by the project, not downstream consumers.

baseline.yaml
```diff
@@ -404,16 +404,17 @@ criteria:
     scorecard_probe:
       - # None, may not be suitable
 
-  - id: OSPS-DO-03
+  - id: OSPS-SA-01
     maturity_level: 2
-    category: Documentation
+    category: Threat Modeling
```
eddie-knight, line 409:
```suggestion
    category: Security Assessment
```

baseline.yaml
```diff
@@ -505,9 +507,9 @@ criteria:
     control_mappings: # TODO
     security_insights_value: # TODO
 
-  - id: OSPS-DO-07
+  - id: OSPS-SA-02
     maturity_level: 2
-    category: Documentation
+    category: Threat Modeling
```
eddie-knight, line 512:
```suggestion
    category: Security Assessment
```

baseline.yaml
```diff
@@ -550,19 +552,19 @@ criteria:
     control_mappings: # TODO
     security_insights_value: # TODO
 
-  - id: OSPS-DO-09
+  - id: OSPS-SA-03
     maturity_level: 3
-    category: Documentation
+    category: Threat Modeling
```
eddie-knight, line 557:
```suggestion
    category: Security Assessment
```

baseline.yaml
```diff
@@ -960,6 +962,32 @@ criteria:
     security_insights_value: # TODO
     scorecard_probe: # sastToolRunsOnAllCommits 
 
+  - id: OSPS-TM-04
+    maturity_level: 2
+    category: Threat Modeling
```
eddie-knight, line 967:
```suggestion
    category: Security Assessment
```

baseline.yaml
evankanderson: Need to either update the "category" label here, or auto-generate it from the `id`.

baseline.yaml
```diff
@@ -960,6 +962,32 @@ criteria:
     security_insights_value: # TODO
     scorecard_probe: # sastToolRunsOnAllCommits 
 
+  - id: OSPS-TM-04
```
eddie-knight, line 965:
```suggestion
  - id: OSPS-SA-04
```

evankanderson:
```suggestion
  - id: OSPS-SA-04
```

baseline.yaml
```diff
@@ -960,6 +962,32 @@ criteria:
...
+      Create a threat model or assessment for the 
+      project. This will identify the assets,
+      threats, and mitigations for the project.
+      Include information on the potential threats
+      to the project's assets and the measures in
+      place to mitigate those threats.
+
+      The authoring process will be simplified if
+      including or referencing the output from
+      OSPS-TM-01, OSPS-TM-02, and OSPS-TM-03.
```
evankanderson, line 987: These links are probably wrong.

evankanderson (commented):
Some concerns about breaking existing linkages.

eddie-knight:
These changes were applied in PR #134

PR #128, opened 2025-01-07
Add criterion involving testing

david-a-wheeler:
Add criteria for automated tests to verify that a changed application still works. Without these, it's not possible to respond in a timely way to vulnerability reports, either in the software or its dependencies, because there's no timely way to verify that the software continues to work correctly.

This is intentionally mirroring OpenSSF Best Practices Badge criteria "test" and "test_policy".

### DIFF
--- a/baseline.yaml
+++ b/baseline.yaml
@@ -125,6 +125,38 @@ criteria:
     control_mappings: # TODO
     scorecard_probe:
       - blocksDeleteOnBranches
+  - id: OSPS-TEST
+    maturity_level: 1
+    category: Build & Release
+    criteria: |
+      The project MUST use at least one automated test suite that
+      is publicly released as OSS (this test suite may be maintained
+      as a separate OSS project). It MUST clearly show
+      or document how to run the test suite(s).
+    objective: |
+      Enable modification of the software and/or its dependencies
+      with reasonable confidence that the changed software will
+      continue to work.
+    implementation: |
+      Establish an automated test suite. Enable it via a
+      continuous integration (CI) script where practical (e.g., where
+      hardware is available),  else via documentation in files such
+      as BUILD.md, README.md, or CONTRIBUTING.md.
+    control_mappings: # TODO
+  - id: OSPS-TEST-ADD
+    maturity_level: 1
+    category: Build & Release
+    criteria: |
+      The project MUST have a general policy (formal or not) that
+      as major new functionality is added to the software produced
+      by the project, tests of that functionality should be added
+      to an automated test suite.
+    objective: |
+      As software is modified its automated test suite should be
+      maintained as well.
+    implementation: |
+      Add automated tests when adding new functionality.
+    control_mappings: # TODO
   - id: OSPS-05
     maturity_level: 1
     category: Build & Release
--- a/baseline.yaml
+++ b/baseline.yaml
@@ -129,10 +129,8 @@ criteria:
     maturity_level: 1
     category: Build & Release
     criteria: |
-      The project MUST use at least one automated test suite that
-      is publicly released as OSS (this test suite may be maintained
-      as a separate OSS project). It MUST clearly show
-      or document how to run the test suite(s).
+      The project MUST use at least one automated test suite for the source
+      code repository which clearly documents when and how it is run.
     objective: |
       Enable modification of the software and/or its dependencies
       with reasonable confidence that the changed software will


david-a-wheeler:
We've talked about these many times, but there doesn't seem to be a matching PR. This at least gets the ball rolling.

baseline.yaml
```diff
@@ -125,6 +125,38 @@ criteria:
...
       - blocksDeleteOnBranches
+  - id: OSPS-TEST
+    maturity_level: 1
+    category: Build & Release
+    criteria: |
+      The project MUST use at least one automated test suite that
+      is publicly released as OSS (this test suite may be maintained
+      as a separate OSS project). It MUST clearly show
+      or document how to run the test suite(s).
+    objective: |
```
funnelfiasco, line 136: It's probably worth calling out in the rationale section that this is primarily driven by being able to more quickly and safely resolve vulnerabilities.

baseline.yaml
```diff
@@ -125,6 +125,38 @@ criteria:
...
+      as BUILD.md, README.md, or CONTRIBUTING.md.
+    control_mappings: # TODO
+  - id: OSPS-TEST-ADD
+    maturity_level: 1
+    category: Build & Release
+    criteria: |
+      The project MUST have a general policy (formal or not) that
+      as major new functionality is added to the software produced
+      by the project, tests of that functionality should be added
+      to an automated test suite.
```
funnelfiasco, line 153: The wording here leaves a lot of wiggle room, to the point that I'm not sure it does very much in practice. Maybe some stronger wording at a higher level?

baseline.yaml
```diff
@@ -125,6 +125,38 @@ criteria:
     control_mappings: # TODO
     scorecard_probe:
       - blocksDeleteOnBranches
+  - id: OSPS-TEST
+    maturity_level: 1
```
funnelfiasco, line 129: I'm going back and forth with myself on the maturity level. Yes, this is unquestionably a Good Idea™ for all projects, but it's also not a practice that's nearly as widely-adopted as we'd like. Is this a criterion that would cause small projects to ignore the rest of the baseline reflexively? I don't know.

funnelfiasco (commented):
This will need updated with the new yaml structure. The first one, I think, is pretty solid, although I'm not sure how I feel about the maturity level. The second one needs wordsmithing, but it's directionally correct.

baseline.yaml
```diff
@@ -125,6 +125,38 @@ criteria:
     control_mappings: # TODO
     scorecard_probe:
       - blocksDeleteOnBranches
+  - id: OSPS-TEST
```
evankanderson, line 128: I'm not sure if this is "quality" or "build and release".  In either case, you probably need a different ID:

```suggestion
  - id: OSPS-QA-08
```

baseline.yaml
evankanderson: I think this is a place where we need to decide to what degree we want to lean on privately-run services like GitHub vs supporting a broad audience.  If we lean on GitHub's services, it's possible for most projects to have at least some pre-merge or post-merge CI.  If not, it becomes hard to measure this.

Interestingly, [OSPS-QA-04](https://baseline.openssf.org/#osps-qa-04) at maturity level 2 requires "any automated status checks for commits" to pass or be manually acknowledged, but there is no requirement that "the project MUST have at least one automated status check applied prior to merging changes".

baseline.yaml
evankanderson: The way I'd generally enforce this would be to leverage test coverage checking, and document a minimum coverage level (for the project as a whole or per-module, depending on my goals).  A link to such coverage measurement might serve as a proxy for validating this criteria, as well?

baseline.yaml
eddie-knight: As highlighted in PR #131, maturity level 1 should work for code or non-code projects. So I think this one has to be at least level 2?

baseline.yaml
```diff
@@ -125,6 +125,38 @@ criteria:
...
+    maturity_level: 1
+    category: Build & Release
+    criteria: |
+      The project MUST use at least one automated test suite that
+      is publicly released as OSS (this test suite may be maintained
+      as a separate OSS project). It MUST clearly show
+      or document how to run the test suite(s).
```
eddie-knight, line 135:
```suggestion
      The project MUST use at least one automated test suite that
      clearly documents when and how it is run.
```

baseline.yaml
eddie-knight, line 153:
```suggestion
      The project documentation MUST include a policy stating that
      new features are to be covered by an automated test suite.
```

baseline.yaml
mlieberman85: Do we define what an automated test suite is? Also, is the expectation here that no one would game this?  For example what would the automated test suite for baseline be?

puerco:
+100 adding a criteria for tests 🎉 
As @david-a-wheeler mentioned during our meeting, setting up the test framework and automation is a big lift but once the project is there, adding new tests means less effort. It also lowers the bar for new contributors to get involved eventually leading to better project sustainability in the long run.

puerco:

PR #130, opened 2025-01-07
Add governance category and mechanism for criteria renaming

### LINKED ISSUE #129: Add "governance" section
It feels like there is probably a "governance" section waiting to happen.  I'd nominate OSPS-DO-01, OSPS-DO-02, OSPS-DO-06, OSPS-DO-11 for inclusion, or it could be combined with the LE- line if desired.

_Originally posted by @evankanderson in https://github.com/ossf/security-baseline/issues/119#issuecomment-2552198644_

evankanderson:
Fixes #129

Per [today's meeting](https://docs.google.com/document/d/16tL1Ln7owIRXSoCKgyYHCs9-JP9iw-ouyk8koGAeHA0/edit?tab=t.0#heading=h.oi20c1w0wg12), re-organize the documentation items relating to management of project contributors into a "governance" category with a "GV" prefix (as "GO" might be mistaken for a programming language).

### DIFF
--- a/baseline.yaml
+++ b/baseline.yaml
@@ -359,12 +359,14 @@ criteria:
     scorecard_probe: # TODO
 
   - id: OSPS-DO-01
+    replaced_by: OSPS-GV-01
+
+  - id: OSPS-GV-01
     maturity_level: 1
-    category: Documentation
+    category: Governance
     criterion: |
       The project MUST have one or more mechanisms
-      for public discussions about proposed
-      changes and usage obstacles.
+      for public discussions about project functionality.
     rationale: |
       Encourages open communication and
       collaboration within the project community,
@@ -372,19 +374,24 @@ criteria:
       discuss proposed changes or usage
       challenges.
     details: |
-      Establish one or more mechanisms for public 
+      Establish and document one or more mechanisms for public 
       discussions within the project, such as
       mailing lists, instant messaging, or issue
       trackers, to facilitate open communication
-      and feedback.
+      and feedback.  The presence of a repository-linked issue tracker,
+      wiki, or a "Feedback" section in the project's README file
+      would meet this criterion.
     control_mappings: # TODO
     security_insights_value: # TODO
     scorecard_probe:
       - # None yet
 
   - id: OSPS-DO-02
+    replaced_by: OSPS-GV-02
+  
+  - id: OSPS-GV-02
     maturity_level: 1
-    category: Documentation
+    category: Governance
     criterion: |
       The project documentation MUST include an
       explanation of the contribution process.
@@ -479,8 +486,11 @@ criteria:
     security_insights_value: # TODO
 
   - id: OSPS-DO-06
+    replaced_by: OSPS-GV-03
+
+  - id: OSPS-GV-03
     maturity_level: 2
-    category: Documentation
+    category: Governance
     criterion: |
       The project documentation MUST include a
       guide for code contributors that includes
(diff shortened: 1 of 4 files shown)


funnelfiasco:
We should probably split this into two:  creating the new category and updating the tooling to support replaced_by, etc. I'm not sure the replaced_by is necessary at this point since we haven't really "published" this, but in either case, these are related but separate concerns.

eddie-knight:
Agreed that this is a great long term value add. Let's not track renames yet, though.

evankanderson:
Do we want to keep track of the "holes" left when we rename criteria?

funnelfiasco:
> Do we want to keep track of the "holes" left when we rename criteria?

My inclination is "no". My take is that we're currently in a pre-release with no API stability promises, so it's okay to make the sort of changes that we wouldn't later.

evankanderson:
> > Do we want to keep track of the "holes" left when we rename criteria?
> 
> My inclination is "no". My take is that we're currently in a pre-release with no API stability promises, so it's okay to make the sort of changes that we wouldn't later.

Do we want to plan then to renumber the existing criteria to collapse the holes that will be left by moving things around?

(Sorry to push on this, but I worry about confusing ourselves as well as other consumers with gaps in the numbering.)

mlieberman85:
I think it's fine pre-release. I do think we want to ensure that post-release we are consistent.

funnelfiasco:
> Do we want to plan then to renumber the existing criteria to collapse the holes that will be left by moving things around?

I was about to say that I don't have a strong preference, but then I realized I think I do. I'm in favor of renumbering, since later on the holes will only be in presentation. (E.g. there won't be an OSPS-XY-123 in the published version, but OSPS-XY-123 will exist in the yaml as a deprecated/renamed/whatever item)

baseline.yaml
```diff
@@ -358,33 +358,34 @@ criteria:
...
       discuss proposed changes or usage
       challenges.
     details: |
-      Establish one or more mechanisms for public 
+      Establish and document one or more mechanisms for public 
       discussions within the project, such as
       mailing lists, instant messaging, or issue
       trackers, to facilitate open communication
-      and feedback.
+      and feedback.  The presence of a repository-linked issue tracker,
```
xee5ch, line 378: I know GitHub and GitLab are the obvious contenders in repository and SCM, but there exists situations with some of these repositories (although niche) where code is public but issue, wiki, or certain documentation not "in the tree" of check-in code is not. Do you want to clarify such example resources _must_ be configured to accept new issues and/or submissions from the public?

xee5ch (commented):
Long-time listener, first time caller. This contribution looks good and the GO-GV category change aligns with [other popular frameworks too](https://nvlpubs.nist.gov/nistpubs/CSWP/NIST.CSWP.29.pdf). I had one minor question or piece of feedback based upon the experiencce of more complex projects that may be worth it, but even if not, these are good changes!

baseline.yaml
mlieberman85: This makes sense though I would maybe limit it to allowing whatever private stuff they want as long as there is a public way to do it. For example, some open source repos do have maintainer only wikis for stuff that is internal and/or sensitive like non-public security vulnerabilities.

evankanderson: I'm willing to update this PR (which is just renumbering and tooling) or open a new PR for that -- the discussion on Friday pointed towards the criteria "there is at least one way for the general public to provide feedback on the software" -- I'm not sure if that included the criteria that the communication channel *also* needed to be visible to other members of the public, but that feels reasonable to me (i.e. a mailing list needs to be public / have public archives to count as such a communication channel).

My preference would be to make that change in a separate PR, to allow this one to merge more quickly.

funnelfiasco: I think just adding the word "public" before "repository-linked" is sufficient to address this concern. If we want to do more substantive wordsmithing, then I agree with Evan that it should be a separate PR so we can land this one.

evankanderson:
Replaced with #134 and #136

PR #137, opened 2025-01-14
Enforce schema when reading YAML

evankanderson:
I noticed that we had some fields (e.g. `implementation`) in the existing criteria that weren't parsed or presented.  Additionally, the format of `control_mappings` differed between the struct (`string`) and the template (`{range .ControlMappings}`).  We don't seem to have used `control_mappings`; I switched it to a list with the idea that it would be easier to switch to a string later if necessary, but a list was probably more correct.

### DIFF
--- a/cmd/baseline.go
+++ b/cmd/baseline.go
@@ -13,14 +13,15 @@ import (
 
 // Struct for representing each entry
 type Criterion struct {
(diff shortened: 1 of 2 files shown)


SecurityCRob (approved):

eddie-knight (approved):
Thanks for this!

PR #140, opened 2025-01-16
Clarify OSPS-BR-01

### LINKED ISSUE #63: Clarify code outside build script

funnelfiasco:
Try to make it better match the original intent. This is another attempt at addressing #63 after discussion in a previous meeting.

Fixes #63

### DIFF
--- a/baseline/OSPS-BR.yaml
+++ b/baseline/OSPS-BR.yaml
@@ -12,17 +12,19 @@ criteria:
     maturity_level: 1
     criterion: |
       The project's build and release pipelines
-      MUST NOT execute arbitrary code that is
-      input from outside of the build script.
+      MUST NOT permit arbitrary input that allows
+      access to privileged resources.
     rationale: |
       Reduce the risk of code injection or other
       security vulnerabilities in the project's
-      build and release processes by restricting
-      the execution of external code.
+      build and release by preventing untrusted input
+      to access privileged resources (code execution,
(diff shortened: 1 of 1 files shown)


SecurityCRob (approved):
i like the rephrasing, ty

eddie-knight (approved):

PR #152, opened 2025-01-17
Split BR-03 into development, release, and consumption requirements.

evankanderson:
BR-03 requires:

> Any websites, API responses or other services involved in the project development and release MUST be delivered using SSH, HTTPS or other encrypted channels

Which, on the face of it, would require that all developers ensure that any email sent relating to the project and project decisions is delivered via SMTPS.  This is probably not intended.

I suspect this criteria will be more clear if we separate it into three concerns:

* Secure access to source code (and use of secure connections during code review) during development.

  This seems well-addressed by git forges such as GitHub, GitLab et al.

* Securing end-user access to released assets, including documentation.  This would align with the best practices badge on [sites_https](https://www.bestpractices.dev/en/criteria?details=true&rationale=true#0.sites_https) and [delivery_mitm](https://www.bestpractices.dev/en/criteria?details=true&rationale=true#0.delivery_mitm).

  This seems addressed by a combination of checking the homepage is HTTPS and using a standard release mechanism (e.g. GitHub releases, language package managers, OCI repositories)

* Securing external content during the release pipeline (e.g. a [SLSA attack D](https://slsa.dev/spec/v1.0/threats-overview) on an insecure download)

  This seems trickier to assess, though we could make a stab at checking Makefiles, shell scripts, and GitHub Actions for `http://` URLs.  This will obviously be leaky.

### DIFF
--- a/baseline/OSPS-BR.yaml
+++ b/baseline/OSPS-BR.yaml
@@ -62,20 +62,19 @@ criteria:
   - id: OSPS-BR-03
     maturity_level: 1
     criterion: |
-      Any websites, API responses or other
-      services involved in the project development
+      Any websites and version control systems
+      involved in the project development
       and release MUST be delivered using SSH,
       HTTPS or other encrypted channels.
     rationale: |
       Protect the confidentiality and integrity
-      of data transmitted between the project's
-      services and users, reducing the risk of
-      eavesdropping or data tampering.
+      of project source code during development,
+      reducing the risk of eavesdropping or data
+      tampering.
     details: |
-      Configure the project's websites, API
-      responses, and other services to use
-      encrypted channels such as SSH or HTTPS for
-      data transmission.
+      Configure the project's websites and version
+      control systems to use encrypted channels
+      such as SSH or HTTPS for data transmission.
     control_mappings: 
       BPB: B-B-11
       CRA: 1.2d, 1.2e, 1.2f, 1.2i, 1.2j, 1.2k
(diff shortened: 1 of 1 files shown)


baseline/OSPS-BR.yaml
```diff
@@ -62,20 +62,19 @@ criteria:
   - id: OSPS-BR-03
     maturity_level: 1
     criterion: |
-      Any websites, API responses or other
-      services involved in the project development
+      Any websites and version control systems
+      involved in the project development
       and release MUST be delivered using SSH,
       HTTPS or other encrypted channels.
```
funnelfiasco, line 68:
```suggestion
      HTTPS, or other encrypted channels.
```

#TeamOxfordComma 😄

baseline/OSPS-BR.yaml
```diff
@@ -62,20 +62,19 @@ criteria:
   - id: OSPS-BR-03
     maturity_level: 1
     criterion: |
-      Any websites, API responses or other
-      services involved in the project development
+      Any websites and version control systems
+      involved in the project development
       and release MUST be delivered using SSH,
```
funnelfiasco, line 67:
```suggestion
      MUST be delivered using SSH,
```

Did you intend to leave "release" in here? It seems like that part is better served in the two new criteria, but I might misunderstand the intent.

funnelfiasco (commented):
I think I'm good with this, other than the question I asked about -03.

baseline/OSPS-BR.yaml
evankanderson: No, 53-character line folding screwed me up.

funnelfiasco (approved):
LGTM

SecurityCRob (approved):
+1

PR #153, opened 2025-01-17
Add suggested implementation for DO-13

evankanderson:
I think we discussed this a bit during the call.

It also appears that the OpenEoX work may not be very stable; I've asked a clarifying question here: https://github.com/oasis-tcs/openeox/issues/64

I also have not looked at _how common_ the different methods are based on [guidance from https://endoflife.date/](https://endoflife.date/recommendations), but that seems like a logical next step.

### DIFF
--- a/baseline/OSPS-DO.yaml
+++ b/baseline/OSPS-DO.yaml
@@ -101,7 +101,14 @@ criteria:
       descriptive statement about the scope and
       duration of support.
     rationale: # TODO
-    implementation: # TODO
+    # TODO: Integrate with advice from https://endoflife.date/recommendations
+    implementation: |
+      The project should have either a "Support"
+      header in the README, a SUPPORT.md file
+      present in the repo root, or a SUPPORT.eox
+      file in the [OpenEOX format](https://github.com/OpenEoX/openeox/blob/main/schema/schema.json)
+      describing the scope and duration of support
+      for the project's released software assets.
     control_mappings: 
       BPB: R-B-3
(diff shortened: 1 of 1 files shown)


funnelfiasco (approved):
I have a weak preference for softening the language a bit to make the methods listed be examples instead of "you must choose from this set" (which may not be the intent, but it's one possible reading). Overall, though, I'm fine with this as-is.

evankanderson:
I'm happy to expand the set -- my real goal was to spark discussion about how to do this.

Looking more at `endoflife.date`, it appears that they have [markdown front-matter to collect this information](https://github.com/endoflife-date/endoflife.date/blob/master/products/istio.md?plain=1), and then [collect this information into JSON files](https://github.com/endoflife-date/release-data/blob/main/releases/istio.json) based on the front-matter and sometimes on [additional scripts](https://github.com/endoflife-date/release-data/tree/main/src).

While this isn't everything, standardizing on something like:

```yaml
auto:
  methods:
  -   git: https://github.com/istio/istio.git
  -   release_table: https://istio.io/latest/docs/releases/supported-releases/
      selector: "table"
      fields:
        releaseCycle: "Version"
        eol:
          column: "End of Life"
          regex: '^~?(?P<value>\w+ \d+(, \d+)?).*'
        supportedKubernetesVersions:
          column: "Supported Kubernetes Versions"
          type: "range"
```

would probably be a nice step forward.

SecurityCRob (approved):
+1

PR #157, opened 2025-01-20
Add links to Minder rule types.

teodor-yanev:
This change adds links to existing Minder rule types that implement Baseline checks.

Note: OSPS-DO-01 and OSPS-DO-02 are no longer included under any of the levels of the security baselines.
Update: Seems like the two rules above have changed their governance criteria as part of https://github.com/ossf/security-baseline/pull/130/files
I will be addressing this separately and adding them in the next batch of rules that we link to Minder -- this will include moving and renaming the files and then updating the yamls again

### DIFF
--- a/baseline/OSPS-AC.yaml
+++ b/baseline/OSPS-AC.yaml
@@ -95,6 +95,9 @@ criteria:
       CSF: PR.AA-02
       OCRE: 486-813, 124-564, 152-725
     security_insights_value: # TODO
+    minder_rules:
+      - name: osps-ac-03
+        url: https://github.com/mindersec/minder-rules-and-profiles/blob/main/security-baseline/rule-types/github/osps-ac-03.yaml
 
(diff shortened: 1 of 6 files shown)


eddie-knight:
I believe when we removed the scorecard values we decided against hard-coupling to other projects, is that right @SecurityCRob? 

Now that I think about it, I'm not sure whether that's applicable to security insights as well.

evankanderson:
As mentioned in the call, we should see about versioning these Minder rules in alignment with the baseline calver versions.

eddie-knight:
Hey team, I'm going to close this PR as part of the decision logged in #189.

This corresponds to #190, which will remove all references to Security Insights.

For the sake of posterity — As a maintainer of Security Insights, I was originally quite optimistic about adding references to implementations, and I'm empathetic to the work that was done in this PR ✊

PR #159, opened 2025-01-24
Updates to Documentation criteria missing rationale and implementation details

SecurityCRob:
<title>

### DIFF
--- a/baseline/OSPS-DO.yaml
+++ b/baseline/OSPS-DO.yaml
@@ -149,26 +149,18 @@ criteria:
       description of how the project selects,
       obtains, and tracks its dependencies.
     rationale: |
-      Providing information about how the project
+      Provide information about how the project
       selects, obtains, and tracks dependencies, 
-      libraries, frameworks, etc. help downstream 
+      libraries, frameworks, etc. to help downstream 
       consumers understand how the project operates
       in regards to third-party components that are
       required neccessary for the software to function.
-
-      This information also helps inform downstream
-      consumer of key third-party components they
-      should be monitoring.
     implementation: |
-      Ideally this information is captured along with the
-      projects technical & design documentation and
-      available through a publicy viewable resource
+      It is recommended to publish this information
+      alongside the projects technical & design
+      documentation on a publicy viewable resource
       such as the source code repository, project website, 
       or other channel.
-
-      Having a "dependencies" header in CONTRIBUTING.md or 
-      equivalent contributing documentation could fulfill
-      this criteria.
     control_mappings: 
       BPB: A-S-1
       CRA: 2.1
--- a/baseline/OSPS-DO.yaml
+++ b/baseline/OSPS-DO.yaml
@@ -101,29 +101,16 @@ criteria:
       descriptive statement about the scope and
       duration of support.
     rationale: |
-      Stating the support expecations the project
-      has (in this case when the project will no
-      longer be actively supporting the project and 
-      considering it "end of life") are important
-      for downstream consumers to be aware of
-      and understand so they can take relevant actions 
-      to ensure the required functionality can still 
-      be achieved.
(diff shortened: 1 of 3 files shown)


baseline/OSPS-DO.yaml
```diff
@@ -100,15 +100,30 @@ criteria:
       The project documentation MUST include a
       descriptive statement about the scope and
       duration of support.
-    rationale: # TODO
-    # TODO: Integrate with advice from https://endoflife.date/recommendations
+    rationale: |
+      Stating the support expecations the project
+      has (in this case when the project will no
+      longer be actively supporting the project and 
+      considering it "end of life") are important
```
evankanderson, line 107: Do we mean the project or versions of a project?

I expect that most projects, like most people, don't expect the end until it happens.

SecurityCRob: hrm....  yes.... both versions and/or project closes up.

baseline/OSPS-DO.yaml
```diff
@@ -100,15 +100,30 @@ criteria:
...
       describing the scope and duration of support
       for the project's released software assets.
+
+      [endoflife.date](https://endoflife.date/recommendations) 
+      provides concrete examples of real-life projects (both good
+      and bad) communicating their end of support.
+
+      Alternatively, the project could published their
+      desired support and duration through such outlets 
+      as a project blog, mailing list, or FAQ.
```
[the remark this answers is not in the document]
SecurityCRob: As much as I desire one location to store this, it has never naturally developed over the last decade(s).  As we roll this out as a standard for LF projects, we could be more prescriptive with how this should be implemented (and automated/templatetize it for our projects) then as others saw it, they could possibly adopt that location.

baseline/OSPS-DO.yaml
```diff
@@ -133,13 +168,28 @@ criteria:
...
+
+      This information also helps inform downstream
+      consumer of key third-party components they
+      should be monitoring.
+    implementation: |
+      Ideally this information is captured along with the
+      projects technical & design documentation and
+      available through a publicy viewable resource
+      such as the source code repository, project website, 
+      or other channel.
```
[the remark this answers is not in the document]
SecurityCRob: great idea!

baseline/OSPS-DO.yaml
evankanderson: Thinking about it... I wouldn't expect (for example) https://github.com/angular/angular to have a "we'll stop supporting the project on $DATE" until the project was already effectively dead.

I _would_ expect them to have a "we'll stop supporting 6.x on $DATE", but that's a more nuanced statement.

funnelfiasco (approved):

baseline/OSPS-DO.yaml
SecurityCRob: I don't want to be prescriptive on what support of comms channels projects must use. With implementation we can enhance the example to help projects understand better what the objective is and how their particular flavour may meet that requirement.  As we get these criteria shored up, we should identify those things that lend themselves to being made into a template, so that folks could easily copy and edit them to meet their needs.

baseline/OSPS-DO.yaml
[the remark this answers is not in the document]
evankanderson: One thing that occurred in discussion this week in Minder was that projects using GitHub releases could actually update existing releases to include some release text indicating when a release is out of support.

I'm not sure what similar options are available for container images, helm charts, and language packages.

eddie-knight (changes requested):
Per our discussion today, I'll move these references and explanations into a lexicon entry for `Duration of Support`

puerco (approved):

eddie-knight (approved):

PR #160, opened 2025-01-24
Update missing rationale & implementation details for Governance criteria

SecurityCRob:
<title>

### DIFF
--- a/baseline/OSPS-GV.yaml
+++ b/baseline/OSPS-GV.yaml
@@ -19,7 +19,7 @@ criteria:
       understand of who is working on the project
       and what areas of authority they may have.
     implementation: |
-      Documenting project particpants and their roles
+      Document project participants and their roles
       through such artifacts as members.md, governance.md, 
       or similar file within the source code repository
       of the project.
--- a/baseline/OSPS-GV.yaml
+++ b/baseline/OSPS-GV.yaml
@@ -17,7 +17,7 @@ criteria:
       helps project particpants, potential contributors,
       and downstream consumers have an accurate
       understand of who is working on the project
-      and what areas or authority they may have.
+      and what areas of authority they may have.
     implementation: |
       Documenting project particpants and their roles
       through such artifacts as members.md, governance.md, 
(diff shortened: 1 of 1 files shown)


baseline/OSPS-GV.yaml
```diff
@@ -12,8 +12,17 @@ criteria:
...
       Roles and Responsibilities for members of the
       project.
-    rationale: # TODO
-    implementation: # TODO
+    rationale: |
+      Documenting project roles and responsibilities
+      helps project particpants, potential contributors,
+      and downstream consumers have an accurate
+      understand of who is working on the project
+      and what areas or authority they may have.
```
funnelfiasco, line 20:
```suggestion
      and what areas of authority they may have.
```
I assume this is what you meant?

baseline/OSPS-GV.yaml
```diff
@@ -12,8 +12,17 @@ criteria:
...
+    implementation: |
+      Documenting project particpants and their roles
```
funnelfiasco, line 22:
```suggestion
      Document project participants and their roles
```
Not sure what our to-be-written style guide will say on this matter, but "Document" seems better than "Documenting" because it's imperative. And either way, the typo fix is good. :-)

funnelfiasco (approved):
Two minor changes that I won't block on, otherwise LGTM

baseline/OSPS-GV.yaml
```diff
@@ -12,8 +12,17 @@ criteria:
...
+      through such artifacts as members.md, governance.md, 
+      or similar file within the source code repository
+      of the project.
```
evankanderson, line 25: What about `MAINTAINERS*` files?  It feels like it would be helpful to both sides to have a list of acceptable file locations, rather than the vague "or similar file with the source code repository".  As an example from elsewhere, we expect `LICENSE*` to be at the top-level, though we allow no-suffix or `.txt` as viable suffixes.

We might be able to use https://codelabs.developers.google.com/codelabs/bigquery-github#0 to get some information as to the prevalence of different files, though that won't cover other locations such as websites.

evankanderson (approved):

baseline/OSPS-GV.yaml
SecurityCRob: oh yeah, forgot that one!  good catch!

evankanderson (approved):

puerco (approved):

PR #161, opened 2025-01-24
Adjust documentation level requirements

evankanderson:
Per discussion in: https://openssf.slack.com/archives/C07DC6TT2QY/p1737570004362569

* Move DO-03 (basic user documentation) to level 1
* Move DO-05 (basic defect reporting) to level 1
* Move DO-13 (end-of-support documentation) to level 2.  There's no clear standard for recording this information at this point.
* Move SA-02 (document APIs) to level 2.  This seems like it should come after DO-03.


SecurityCRob (approved):
i'm fine with these level adjustments

PR #163, opened 2025-01-24
Add dependency/SBOM criteria and releveling

puerco:
This commit updates the dependency criteria to add increasing transparency requirements at each level:

1. Checked in language-native dependency list 
2. Simple SBOM
3. SBOM with NTIA/CRA data fields

In addition to the two new ones, OSPS-QA-03 is simplified and releveled to 1.

### DIFF
--- a/baseline/OSPS-QA.yaml
+++ b/baseline/OSPS-QA.yaml
@@ -232,60 +232,22 @@ criteria:
     maturity_level: 3
     criterion: |
       All released built software assets MUST be
-      delivered with a software bill of materials
-      (SBOM) in SPDX or CycloneDX format accounting
-      all direct software dependencies with their
-      associated version identifiers.
+      delivered with a software bill of materials.
     rationale: |
       Provide transparency and accountability for
-      the project's dependencies in a standard format
-      allowing automated systems to understand the 
+      the project's dependencies in a standard format,
+      allowing automated systems to understand the
       software's dependencies and versions.
     details: |
-      It is recommended to use a CycloneDX or SPDX
-      file that is auto-generated at build time by
-      a tool that has been vetted for accuracy.
+      It is recommended to auto-generate SBOMs at build time
+      using a tool that has been vetted for accuracy.
       This enables users to ingest this data in a
       standardized approach alongside other
       projects in their environment.
-    control_mappings: 
+    control_mappings:
       BPB: Q-S-9
       CRA: 13.24, I.II.1, VII.2
       SSDF: PS3, PW4
       CSF: ID.AM-02
       OC: 4.3.1
       OCRE: 486-813, 124-564, 863-521
-    security_insights_value: # TODO
-
-  - id: OSPS-QA-12
-    maturity_level: 3
-    criterion: |
-      The project will provide both source and build software bills 
-      of materials the following data elements for the components 
-      included in the build: 
-      - license 
-      - supplier name 
-      - filename of the component
-      - component name
-      - component version 
-      - software identifiers  
-      - relationship between the components 
-      - author of the SBOM data and timestamp  
-      
-      Additionally, for deployable and executable components
-      the SBOM should record their cryptographic hashes.
-    rationale: |
-      Provide the minimal data elements required to validate,
-      verify the integrity of software components and assess risk.
-    details: |
-      SBOM generation tools can specialize on one kind of data element.
-      Supplying the required data in multiple software bills of materials 
-      can be done although, ideally, data should be combined into one.
-    control_mappings: 
-      BPB: Q-S-9
-      CRA: 13.24, I.II.1, II.2
-      SSDF: PS3, PW4
-      CSF: ID.AM-02
-      OC: 4.3.1
-      OCRE: 486-813, 124-564, 863-521
-    security_insights_value: # TODO
--- a/baseline/OSPS-QA.yaml
+++ b/baseline/OSPS-QA.yaml
@@ -80,10 +80,10 @@ criteria:
     control_mappings: 
       BPB: Q-S-8, Q-S-9
       CRA: 2.1, 2.3
-      SSDF: PO3.3, PS1
+      SSDF: PO3.3, PS1, PS3.2
       CSF: ID.AM-02
       OC: 4.1.5, 4.3.1
-      OCRE: 486-813, 124-564, 673-475,863-521
+      OCRE: 486-813, 124-564, 673-475,863-521, 613-286
     security_insights_value: # TODO
 
   - id: OSPS-QA-04
@@ -244,10 +244,11 @@ criteria:
       This enables users to ingest this data in a
       standardized approach alongside other
       projects in their environment.
-    control_mappings:
-      BPB: Q-S-9
-      CRA: 13.24, I.II.1, VII.2
-      SSDF: PS3, PW4
-      CSF: ID.AM-02
+    control_mappings: 
+      BPB: Q-S-8, A-S-1
+      CRA: 2.1
+      SSDF: PS3.2, PW4
+      CSF: ID.AM-01, ID.AM-02
       OC: 4.3.1
-      OCRE: 486-813, 124-564, 863-521
+      OCRE: 486-813, 124-564, 863-521, 613-286
+    security_insights_value: # TODO
--- a/baseline/OSPS-QA.yaml
+++ b/baseline/OSPS-QA.yaml
@@ -55,7 +55,7 @@ criteria:
     control_mappings: 
       BPB: CC-B-2, CC-B-3, R-B-5
       CRA: 1.2b, 1.2f, 1.2j
-      SSDF: PO3.2, PS1, PS2, PS3, PW1.2, PW2.1,  
+      SSDF: PO3.2, PS1, PS2, PS3, PW1.2, PW2.1
       CSF: ID.AM-02, ID.RA-01, ID.RA-08
       OC: 4.1.4
       OCRE: 486-813, 124-564, 757-271
--- a/.github/ISSUE_TEMPLATE/release.yaml
+++ b/.github/ISSUE_TEMPLATE/release.yaml
@@ -1,0 +1,33 @@
+name: New release
+description: Checklist for creating a new release of the OSPS Baseline
+title: "[Release]: "
+body:
+  - type: markdown
+    attributes:
+      value: |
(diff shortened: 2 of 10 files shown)


baseline/OSPS-QA.yaml
```diff
@@ -62,31 +62,21 @@ criteria:
...
+    maturity_level: 1
     criterion: |
-      All released software assets MUST be
-      delivered with a machine-readable list of
-      all direct and transitive internal software
-      dependencies with their associated version
-      identifiers.
+      The code repository should contain a dependency 
+      list that accounts for the direct language dependencies
+      when the package management system supports it.
```
evankanderson, line 69: We haven't written anything tracking this requirement yet, but rewriting these criteria between RCs will be more of a challenge to coordinate in the future as we start measuring pilot projects against them.

(Fine for now)

SecurityCRob (approved):
i need to adjust the xls to account for this change, but +1

baseline/OSPS-QA.yaml
```diff
@@ -62,31 +62,21 @@ criteria:
...
-
-      It is recommended to use a CycloneDX or SPDX
-      file that is auto-generated at build time by
-      a tool that has been vetted for accuracy.
-      This enables users to ingest this data in a
-      standardized approach alongside other
-      projects in their environment.
+      This may take the form a package manager or
+      language dependency lock file that ennumerates all
+      direct and transitive dependencies such as
```
evankanderson, line 78: Details says "direct and transitive" but criteria says "direct".  Which do we require here?

baseline/OSPS-QA.yaml
```diff
@@ -237,3 +227,60 @@ criteria:
...
       BPB: B-G-3
     security_insights_value: # TODO
+
+  - id: OSPS-QA-11
+    maturity_level: 2
+    criterion: |
+      All released software assets MUST be
+      delivered with a software bill of materials
+      (SBOM) in SPDX or CycloneDX format accounting
+      all direct software dependencies with their
```
evankanderson, line 237: Do we mean only direct dependencies here, or also transitive dependencies for this level-2 criteria?

evankanderson: (We don't say anything about transitive dependencies except in the details on QA-03)

puerco: the original QA-03 required direct and transitive. I dropped it to require only  the direct dependencies as both NTIA and BSI only require just the direct ones.

puerco:
> i need to adjust the xls to account for this change,

@SecurityCRob I'm happy to add it to the spreadsheet.
Note that OSPS-QA-11 & OSPS-QA-12 match SSDF PS3, PW4.   OSPS-QA-03 is cross referenced to SSDF PO4 and PS1. I think those are not correct, but I didn't want to mess with your original assessment.

baseline/OSPS-QA.yaml
```diff
@@ -62,31 +62,21 @@ criteria:
...
   - id: OSPS-QA-03
-    maturity_level: 2
+    maturity_level: 1
     criterion: |
-      All released software assets MUST be
-      delivered with a machine-readable list of
-      all direct and transitive internal software
-      dependencies with their associated version
-      identifiers.
+      The code repository should contain a dependency
```
eddie-knight, line 67:
```suggestion
      The code repository MUST contain a dependency 
```

eddie-knight (changes requested):
I noticed that the first entry needs to be changed to follow our criterion format

baseline/OSPS-QA.yaml
puerco: Done :+1:

baseline/OSPS-QA.yaml
```diff
@@ -62,31 +62,21 @@ criteria:
...
-      It is recommended to use a CycloneDX or SPDX
-      file that is auto-generated at build time by
-      a tool that has been vetted for accuracy.
-      This enables users to ingest this data in a
-      standardized approach alongside other
-      projects in their environment.
+      This may take the form a package manager or
+      language dependency lock file that ennumerates all
+      direct and, ideally, transitive build dependencies such as
+      package.json, Gemfile.lock, or go.sum.
```
evankanderson, line 79: Do we include `requirements.txt` here despite the fact that it is not a lockfile, or no?

I don't have a strong preference, except that I'm going to write a policy to check this.

evankanderson (approved):

baseline/OSPS-QA.yaml
funnelfiasco: `requirements.txt` _could_ be used to satisfy it as written, but doesn't necessarily. I do think there's a mismatch between the criterion and the rationale/details that I will raise as a separate suggestion.

baseline/OSPS-QA.yaml
```diff
@@ -62,31 +62,21 @@ criteria:
...
-      identifiers.
+      The code repository MUST contain a dependency 
+      list that accounts for the direct language dependencies
+      when the package management system supports it.
     rationale: |
       Provide transparency and accountability for
-      the project's dependencies, enabling users
-      and contributors to understand the
-      software's dependencies and versions.
+      the project's dependencies by fixing the desired versions
```
funnelfiasco, line 72:
```suggestion
      the project's dependencies
```

For level 1, I'm not sure we want to require version pinning (or if we do, we need to be very clear about what is a sufficient strictness). To my mind, listing the dependencies is sufficient for level 1.

baseline/OSPS-QA.yaml
```diff
@@ -62,31 +62,21 @@ criteria:
...
-      go.sum.
-
-      It is recommended to use a CycloneDX or SPDX
-      file that is auto-generated at build time by
-      a tool that has been vetted for accuracy.
-      This enables users to ingest this data in a
-      standardized approach alongside other
-      projects in their environment.
+      This may take the form a package manager or
+      language dependency lock file that ennumerates all
```
funnelfiasco, line 77:
```suggestion
      language dependency file that ennumerates all
```

funnelfiasco (changes requested):
I think we should soften the language in the rationale/details to make it clear that what we're looking for in this criterion is enumeration, not pinning.

On the other hand, if we _do_ want to require some form of pinning at level 1, we need to
1. Update the criterion text to include that
2. Specify what sort of pinning is sufficient (e.g. minimum version, branch/release name, specific version, hash, etc etc etc)

For clarity: my strong preference is to go with the first approach and focus on enumeration at level 1.

SecurityCRob:
> > i need to adjust the xls to account for this change,
> 
> @SecurityCRob I'm happy to add it to the spreadsheet. Note that OSPS-QA-11 & OSPS-QA-12 match SSDF PS3, PW4. OSPS-QA-03 is cross referenced to SSDF PO4 and PS1. I think those are not correct, but I didn't want to mess with your original assessment.

I'll figure this out today, no worries!

funnelfiasco (approved):

baseline/OSPS-QA.yaml
```diff
@@ -62,31 +62,21 @@ criteria:
...
-      It is recommended to use a CycloneDX or SPDX
-      file that is auto-generated at build time by
-      a tool that has been vetted for accuracy.
-      This enables users to ingest this data in a
-      standardized approach alongside other
-      projects in their environment.
+      This may take the form a package manager or
+      language dependency file that ennumerates all
+      direct and, ideally, transitive build dependencies such as
+      package.json, Gemfile.lock, or go.sum.
```
eleftherias, line 79:
```suggestion
      package-lock.json, Gemfile.lock, or go.sum.
```
I imagine the Javascript file that satisfies this criterion would be a `package-lock.json` or `yarn.lock`, instead of the `package.json`. The `package.json` file is similar to the `go.mod` file, not `go.sum`.

baseline/OSPS-QA.yaml
```diff
@@ -237,3 +227,65 @@ criteria:
     control_mappings:       
       BPB: B-G-3
     security_insights_value: # TODO
+
+  - id: OSPS-QA-11
+    maturity_level: 2
+    criterion: |
+      All released software assets MUST be
```
david-a-wheeler, line 234:
```suggestion
      All released built software assets MUST be
```

The rest of the text presumes it's built, but the criterion doesn't say so.

baseline/OSPS-QA.yaml
```diff
@@ -237,3 +227,65 @@ criteria:
     control_mappings:       
       BPB: B-G-3
     security_insights_value: # TODO
+
+  - id: OSPS-QA-11
+    maturity_level: 2
```
david-a-wheeler, line 232:
```suggestion
    maturity_level: 3
```

I'd suggest this as a 3. This is not a common practice, and in many situations these can be derived.

baseline/OSPS-QA.yaml
```diff
@@ -237,3 +227,65 @@ criteria:
...
+      OC: 4.3.1
+      OCRE: 486-813, 124-564, 863-521
+    security_insights_value: # TODO
+
+  - id: OSPS-QA-12
+    maturity_level: 3
+    criterion: |
+      The project will provide both source and build software bills 
+      of materials the following data elements for the components 
+      used in the build:
```
david-a-wheeler, line 265:
```suggestion
      included in the build: 
```

Few should care about the platform used to run a build. Knowing the tools used to generate & test are of interest, but let's walk before we run.

baseline/OSPS-QA.yaml
```diff
@@ -62,31 +62,21 @@ criteria:
...
-      It is recommended to use a CycloneDX or SPDX
-      file that is auto-generated at build time by
-      a tool that has been vetted for accuracy.
-      This enables users to ingest this data in a
-      standardized approach alongside other
-      projects in their environment.
+      This may take the form a package manager or
+      language dependency file that ennumerates all
+      direct and, ideally, transitive build dependencies such as
+      package-lock.json, Gemfile.lock, or go.sum.
```
eleftherias, line 79: The way this is currently written, the rationale and details sections seem to describe different things. Using Go as an example, the rationale section describes having a `go.mod` file which the details section describes a `go.sum`.

Given that lockfiles aren't required or consistently used in some languages, like Python or Java, 
I suggest making this criterion about enumeration only (`go.mod`) since it's a generic enough requirement to apply to all languages. 

```suggestion
      This may take the form a package manager or
      language dependency file that ennumerates all
      direct dependencies such as package.json, 
      Gemfile, or go.mod.
```

PR #158, opened 2025-01-24
Added lexicon entry for CRA

SecurityCRob:
test run of adding lexicon entries for reg/framework mappings and references using the CRA as our 1st entry

PR #164, opened 2025-01-24
Added template support for lexicon references

eddie-knight:
follows PR #158

### DIFF
--- a/cmd/template.md
+++ b/cmd/template.md
@@ -9,6 +9,11 @@ The Open Source Project Security (OSPS) Baseline is a set of security criteria t
 The criteria are organized by maturity level and category.
 In the detailed subsections you will find the criterion, rationale, and details notes.
 
+
+Where possible, we have added control mappings to external frameworks.
+These are not guaranteed to be 100% matches, but instead serve as references
+when working to meet the corresponding controls.
+
 For more information on the project and to make contributions, visit the [GitHub repo](https://github.com/ossf/security-baseline).
 
 ---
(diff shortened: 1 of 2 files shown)


evankanderson (approved):

cmd/template.md
```diff
@@ -88,6 +88,18 @@ For more information on the project and to make contributions, visit the [GitHub
 ### {{ .Term }}
 
 {{ .Definition }}
+{{ if .Synonyms }}
+{{- $length := len .Synonyms -}}
+**Synonyms:** _{{- range $index, $synonym := .Synonyms -}}{{ $synonym }}{{- if lt $index (subtract $length 1) }}, {{ end -}}
+{{- end -}}_
+{{- end }}
```
funnelfiasco, line 95: Do we want this here or do we want to just leave them as links at the bottom as they are currently? I don't have a strong preference either way.

funnelfiasco (approved):

eddie-knight:
Edit: Working on the preview, and getting some unexpected behavior when built from GitHub Actions vs locally. Working on it.

cmd/template.md
eddie-knight: Unless I'm missing/misunderstanding— right now these values are only visible in the YAML and not the markdown

eddie-knight: After staring at them for a while, I didn't really like how these look either— almost all of the synonyms are just for plurals. Let's omit and make sure that any info in the Synonyms field doesn't need to be included in the markdown.

PR #169, opened 2025-01-27
Update BR category to OSPS-BR-xxx numbering

SecurityCRob:
adjusting criterias' number to "XXX" format

### DIFF
--- a/baseline/OSPS-BR.yaml
+++ b/baseline/OSPS-BR.yaml
@@ -8,7 +8,7 @@ description: |
   vulnerabilities or errors in the software distribution
   process.
 criteria:
-  - id: OSPS-BR-01
+  - id: OSPS-BR-101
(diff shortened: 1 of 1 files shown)


funnelfiasco:
If we're going to do this, and I'm not convinced we should, we should have them be in numerical order in the yaml file. Otherwise, it's going to be hard to see which is the next number to be used.

eddie-knight (approved):

SecurityCRob:
> If we're going to do this, and I'm not convinced we should, we should have them be in numerical order in the yaml file. Otherwise, it's going to be hard to see which is the next number to be used.

very much agree, but didn't want to make too many changes without the group debating first.  If we want to go this route I'll update all the numbering pr's

eddie-knight:

PR #170, opened 2025-01-27
Update DO category to OSPS-DO-xxx format

SecurityCRob:
swap DO from 2digit ids "XX"  to 3digit "XXX"
fix one miscategorization


funnelfiasco:
I'll repeat what I said in #169:
> If we're going to do this, and I'm not convinced we should, we should have them be in numerical order in the yaml file. Otherwise, it's going to be hard to see which is the next number to be used.

eddie-knight (approved):

PR #171, opened 2025-01-27
Update GV category to OSPS-GV-xxx numbering

SecurityCRob:
going from 2digits to 3


eddie-knight:

PR #173, opened 2025-01-27
Update SA category to OSPS-SA-xxx numbering

SecurityCRob:
2 to 3digit ids


eddie-knight:

PR #191, opened 2025-02-18
Update absolute phrasing "considered secure"

jkjell:
I've previously been advised using the phrase "secure" without qualification can be... problematic. ☺️


funnelfiasco (approved):
This is wise.

eddie-knight (approved):
good looking out! thanks @jkjell

PR #194, opened 2025-02-19
Boostrap oscal generator and subcommand

puerco:
This command initializes the first oscal output and generator for the base line data.

This lets us generate a first draft of an OSCAL catalog to start testing and iterating on the format.

Here's a sample of the output for feedback
https://gist.github.com/puerco/e4a18354c281d45fc900cfb626adf286

/cc @brandtkeller 

### DIFF
--- a/cmd/pkg/baseline/generator_oscal.go
+++ b/cmd/pkg/baseline/generator_oscal.go
@@ -18,6 +18,9 @@ const (
 	VersionOSPS = "devel"
 	controlHREF = "https://baseline.openssf.org/versions/%s#%s"
 	catalogUUID = "8c222a23-fc7e-4ad8-b6dd-289014f07a9f"
+
+	// OpenSSFNS is the OSCAL namespace URI to define the baseline names.
+	OpenSSFNS = "http://baseline.openssf.org/ns/oscal"
 )
 
 func (g *Generator) ExportOSCAL(b *types.Baseline, w io.Writer) error {
@@ -30,6 +33,7 @@ func (g *Generator) ExportOSCAL(b *types.Baseline, w io.Writer) error {
 			Links: &[]oscal.Link{
 				{
 					Href: fmt.Sprintf(controlHREF, VersionOSPS, ""),
+					Rel:  "canonical",
 				},
 			},
 			OscalVersion: "1.1.3",
@@ -58,13 +62,20 @@ func (g *Generator) ExportOSCAL(b *types.Baseline, w io.Writer) error {
 				parts = append(parts, oscal.Part{
 					Class: control.ID,
 					ID:    ar.ID,
-					Links: &[]oscal.Link{},
-					Name:  "",
+					Name:  ar.ID,
 					Ns:    "",
 					Parts: &[]oscal.Part{
 						{
-							ID:    ar.ID + "_recommendation",
+							ID:    ar.ID + ".R",
+							Name:  "recomemendation",
+							Ns:    OpenSSFNS,
 							Prose: ar.Recommendation,
+							Links: &[]oscal.Link{
+								{
+									Href: fmt.Sprintf(controlHREF, VersionOSPS, ar.ID),
+									Rel:  "canonical",
+								},
+							},
 						},
 					},
 					Prose: ar.Text,
@@ -73,12 +84,12 @@ func (g *Generator) ExportOSCAL(b *types.Baseline, w io.Writer) error {
 			}
 
 			newCtl := oscal.Control{
-				Class: control.ID[0:7], // OSPS-BR-01,
+				Class: code,
 				ID:    control.ID,
 				Links: &[]oscal.Link{
 					{
 						Href: fmt.Sprintf(controlHREF, VersionOSPS, strings.ToLower(control.ID)),
-						Rel:  "reference",
+						Rel:  "canonical",
 					},
 				},
 				Parts: &parts,
@@ -92,9 +103,17 @@ func (g *Generator) ExportOSCAL(b *types.Baseline, w io.Writer) error {
 	}
 	catalog.Groups = &catalogGroups
 
+	// Wrap the catalog to render the required "catalog" wrapper
+	// in the JSON file:
+	var wrapper = struct {
+		Catalog oscal.Catalog `json:"catalog"`
+	}{
+		Catalog: catalog,
+	}
+
 	enc := json.NewEncoder(w)
 	enc.SetIndent("", "  ")
-	if err := enc.Encode(catalog); err != nil {
+	if err := enc.Encode(wrapper); err != nil {
 		return fmt.Errorf("encoding oscal json data: %w", err)
 	}
 	return nil
--- a/cmd/pkg/baseline/generator_oscal.go
+++ b/cmd/pkg/baseline/generator_oscal.go
@@ -1,0 +1,107 @@
+// SPDX-FileCopyrightText: Copyright 2025 The OSPS Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package baseline
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"strings"
+	"time"
+
+	oscal "github.com/defenseunicorns/go-oscal/src/types/oscal-1-1-3"
+	"github.com/ossf/security-baseline/pkg/types"
+)
+
+const (
+	VersionOSPS = "devel"
+	controlHREF = "https://baseline.openssf.org/versions/%s#%s"
+	catalogUUID = "8c222a23-fc7e-4ad8-b6dd-289014f07a9f"
+)
+
+func (g *Generator) ExportOSCAL(b *types.Baseline, w io.Writer) error {
+	n := time.Now()
+	catalog := oscal.Catalog{
+		UUID:   catalogUUID,
+		Groups: nil,
+		Metadata: oscal.Metadata{
+			LastModified: n,
+			Links: &[]oscal.Link{
+				{
+					Href: fmt.Sprintf(controlHREF, VersionOSPS, ""),
+				},
+			},
+			OscalVersion: "1.1.3",
+			Published:    &n,
+			Title:        "Open Source Project Security Baseline",
+			Version:      VersionOSPS,
+		},
+	}
+
+	catalogGroups := []oscal.Group{}
+
+	for code, cat := range b.Categories {
+		group := oscal.Group{
+			Class:    "OSPS",
+			Controls: nil,
+			ID:       code,
+			Title:    cat.Description,
+		}
+
+		controls := []oscal.Control{}
+
+		for _, criteria := range cat.Criteria {
+			newCtl := oscal.Control{
+				Class: criteria.Category,
+				ID:    strings.TrimPrefix(criteria.ID, "OSPS-"),
+				Links: &[]oscal.Link{
+					{
+						Href: fmt.Sprintf(controlHREF, VersionOSPS, strings.ToLower(criteria.ID)),
+						Rel:  "reference",
+					},
+				},
+				Parts: &[]oscal.Part{
+					{
+						ID:    strings.TrimPrefix(criteria.ID, "OSPS-") + "_level",
+						Name:  "maturity-level",
+						Prose: fmt.Sprintf("%d", criteria.MaturityLevel),
+					},
+					{
+						ID:    strings.TrimPrefix(criteria.ID, "OSPS-") + "_details",
+						Name:  "details",
+						Prose: strings.TrimSpace(criteria.Details),
+					},
+					{
+						ID:    strings.TrimPrefix(criteria.ID, "OSPS-") + "_rationale",
+						Name:  "rationale",
+						Prose: strings.TrimSpace(criteria.Rationale),
+					},
+				},
+				Title: strings.TrimSpace(criteria.CriterionText),
+			}
+
+			if criteria.Implementation != "" {
+				pts := append(*newCtl.Parts, oscal.Part{
+					ID:    strings.TrimPrefix(criteria.ID, "OSPS-") + "_implementation",
+					Name:  "implementation",
+					Prose: strings.TrimSpace(criteria.Implementation),
+				})
+				newCtl.Parts = &pts
+			}
+
+			controls = append(controls, newCtl)
+		}
+
+		group.Controls = &controls
+		catalogGroups = append(catalogGroups, group)
+	}
+	catalog.Groups = &catalogGroups
+
+	enc := json.NewEncoder(w)
+	enc.SetIndent("", "  ")
+	if err := enc.Encode(catalog); err != nil {
+		return fmt.Errorf("encoding oscal json data: %w", err)
+	}
+	return nil
+}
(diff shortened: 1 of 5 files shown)


cmd/pkg/baseline/generator_oscal.go
```diff
@@ -0,0 +1,107 @@
...
+						Href: fmt.Sprintf(controlHREF, VersionOSPS, strings.ToLower(criteria.ID)),
+						Rel:  "reference",
+					},
+				},
+				Parts: &[]oscal.Part{
+					{
+						ID:    strings.TrimPrefix(criteria.ID, "OSPS-") + "_level",
+						Name:  "maturity-level",
+						Prose: fmt.Sprintf("%d", criteria.MaturityLevel),
+					},
```
xee5ch, line 69: I would strongly recommend using OSCAL `prop`s (`properties: []` in the JSON data format) over parts for this.

xee5ch (changes requested):

cmd/pkg/baseline/generator_oscal.go
puerco: Thanks for the feedback, love the suggestion! I've pushed a new commit moving maturity level to a property. Check out the new output here
https://gist.github.com/puerco/e4a18354c281d45fc900cfb626adf286

cmd/pkg/baseline/generator_oscal.go
```diff
@@ -0,0 +1,109 @@
...
+		}
+
+		group.Controls = &controls
+		catalogGroups = append(catalogGroups, group)
+	}
+	catalog.Groups = &catalogGroups
+
+	enc := json.NewEncoder(w)
+	enc.SetIndent("", "  ")
+	if err := enc.Encode(catalog); err != nil {
```
brandtkeller, line 105: This will produce the file for all of the content within the catalog model - but if we look at a [catalog example](https://github.com/usnistgov/oscal-content/blob/main/examples/catalog/json/basic-catalog.json#L2) we need the `catalog` object key included for validity of the model. 

This can be done a number of ways. Typically we use the [complete schema](https://github.com/defenseunicorns/go-oscal/blob/v0.6.2/src/types/oscal-1-1-3/types.go#L23) to reference the catalog and encode the complete schema. 

go-oscal has a CLI with a `validate` command that can perform schema validation of models if that is helpful.

cmd/pkg/baseline/generator_oscal.go
```diff
@@ -0,0 +1,109 @@
...
+	"io"
+	"strings"
+	"time"
+
+	oscal "github.com/defenseunicorns/go-oscal/src/types/oscal-1-1-3"
+	"github.com/ossf/security-baseline/pkg/types"
+)
+
+const (
+	VersionOSPS = "devel"
```
funnelfiasco, line 18: So how will this work when we release a version? Currently, the SOP is to copy the rendered .md to a directory. How do we generate the OSCAL JSON with the correct version strings?

xee5ch: Is just "this current version" the issue at hand, @funnelfiasco, or you want to consider how you manage "the lifecycle of versions and say where this version is a series of versions" kind of perspective? OSCAL can encode both, not sure how you all want to handle it.

cmd/pkg/baseline/generator_oscal.go
```diff
@@ -0,0 +1,109 @@
...
+		for _, criteria := range cat.Criteria {
+			newCtl := oscal.Control{
+				Class: criteria.Category,
+				ID:    strings.TrimPrefix(criteria.ID, "OSPS-"),
+				Links: &[]oscal.Link{
+					{
+						Href: fmt.Sprintf(controlHREF, VersionOSPS, strings.ToLower(criteria.ID)),
+						Rel:  "reference",
+					},
+				},
```
xee5ch, line 63: Can you explain what the goal of the references are, aside from the OSCAL data format, what are we trying to accomplish with them in the baseline? I want to make sure I understand before I provide possibly another recommendation.

It will be stylistic, this is all syntactically valid, I want to understand the goal.

cmd/pkg/baseline/generator_oscal.go
funnelfiasco: My concern is that including a link to the "devel" version of a control may not give the right information if the file was generated against a previous version of the OSPS Baseline. It's not clear to me what @puerco intends to do with the generated json. Is it something we'll publish alongside the versioned baseline files? Is it something that we expect people to generate ad hoc if they need it? Something else?

xee5ch: OK cool. So there is simple and complex ways to handle this in OSCAL, I am not sure which one suits the group best. Each OSCAL model, of course the beloved catalog, uses the same `//metadata` assembly and that can have zero, one or many [`revisions`](https://pages.nist.gov/OSCAL-Reference/models/v1.1.3/catalog/json-reference/#/catalog/metadata/revisions). Now you could "reinvent all of git" (and I say that sarcastically, who would do that?) and enumerate every revision. But there is another thing you could do here, if it suits you all. You see the docs talk about this:

> Constraint (1)
>
> allowed values for link/@rel
>
> The value may be locally defined, or one of the following:
>
>    canonical: The link identifies the authoritative location for this resource. Defined by RFC 6596.
>    alternate: The link identifies an alternative location or format for this resource. Defined by the HTML Living Standard
>    predecessor-version: This link identifies a resource containing the predecessor version in the version history. Defined by RFC 5829.
>    successor-version: This link identifies a resource containing the predecessor version in the version history. Defined by RFC 5829.
>    version-history: This link identifies a resource containing the version history of this document. Defined by RFC 5829.

So why do I bring this up? You could decide to keep the version at `devel` for now or figure out some other ID. And you can point to the `version-history` to say where all releases are, before or after you release them. Something like so:

```yaml
---
catalog:
  uuid: 394c88cf-07c0-4380-899b-053810d8b21e
  version: devel # or maybe something different?
  oscal-version: 1.1.3
  last-modified: 2025-02-20T03:43:57Z
  published:      2025-02-20T03:43:57Z
  metadata:
    revisions:
    - title: OpenSSF Security Baseline Releases
      links:
      - rel: version-history
        href: https://github.com/ossf/security-baseline/releases
```

You could also do something different with `predecessor-version` or `successor-version` once you have a release or two under your belt, but I find that more complicated than this approach with that `@rel` type. That's why it got added. As for `devel` there are some other things you could _try_ but I will say my goto that is problematic is a git SHA1 hash? Why not that? Well, because if you want to hash the content and make that the version of the catalog before a release, it is kind of a recursive hell where you need to answer "how do I embed the hash of the content into the content I am hashing?" so I am out of good pre-release ideas, but this facility does give you what you need for the "link to previous versions of the baseline" scenario maybe?

cmd/pkg/baseline/generator_oscal.go
```diff
@@ -0,0 +1,109 @@
...
+				Props: &[]oscal.Property{
+					{
+						Name:  "maturity-level",
+						UUID:  catalogUUID,
+						Value: fmt.Sprintf("%d", criteria.MaturityLevel),
+					},
```
xee5ch, line 69: I like the update here to use `prop`, but another minor recommendation given [extending and customizing data](https://pages.nist.gov/OSCAL/learn/tutorials/general/extension/), it might make sense to namespace (with `ns`) that `maturity-level` has a specific meaning in the context of OpenSSF and if other organizations have a similar prop, they do not "collide" with its use or downstream code assumes some other group's use of the term applies here. The go-oscal library seems to support this, so I'd recommend picking a ns URL to use in props and other things and sticking to it through the codebase for namespacing per the guidance.

I am not saying it has to be that specific value, but just giving you an idea of where I am going with it.

```suggestion
				Props: &[]oscal.Property{
					{
					        Ns: "http://openssf.org/ns/oscal",
						Name:  "maturity-level",
						UUID:  catalogUUID,
						Value: fmt.Sprintf("%d", criteria.MaturityLevel),
					},
```

xee5ch (changes requested):

cmd/pkg/baseline/generator_oscal.go
```diff
@@ -0,0 +1,109 @@
...
+				},
+				Parts: &[]oscal.Part{
+					{
+						ID:    strings.TrimPrefix(criteria.ID, "OSPS-") + "_details",
+						Name:  "details",
+						Prose: strings.TrimSpace(criteria.Details),
+					},
```
xee5ch, line 76: I would recommend [per the OSCAL docs on `/catalog/parts/@name`](https://pages.nist.gov/OSCAL-Reference/models/v1.1.3/catalog/json-reference/#/catalog/groups/controls/parts) to change this to `statement`.

```suggestion
					{
						ID:    strings.TrimPrefix(criteria.ID, "OSPS-") + "_statement",
						Name:  "statement",
						Prose: strings.TrimSpace(criteria.Details),
					},
```

cmd/pkg/baseline/generator_oscal.go
```diff
@@ -0,0 +1,109 @@
...
+					{
+						ID:    strings.TrimPrefix(criteria.ID, "OSPS-") + "_rationale",
+						Name:  "rationale",
+						Prose: strings.TrimSpace(criteria.Rationale),
+					},
```
xee5ch, line 81: I would recommend [per the OSCAL docs on `/catalog/parts/@name`](https://pages.nist.gov/OSCAL-Reference/models/v1.1.3/catalog/json-reference/#/catalog/groups/controls/parts) to change this to `guidance`.

```suggestion
					{
						ID:    strings.TrimPrefix(criteria.ID, "OSPS-") + "_guidance",
						Name:  "guidance",
						Prose: strings.TrimSpace(criteria.Rationale),
					},
```

cmd/pkg/baseline/generator_oscal.go
```diff
@@ -0,0 +1,109 @@
...
+				},
+				Title: strings.TrimSpace(criteria.CriterionText),
+			}
+
+			if criteria.Implementation != "" {
+				pts := append(*newCtl.Parts, oscal.Part{
+					ID:    strings.TrimPrefix(criteria.ID, "OSPS-") + "_implementation",
+					Name:  "implementation",
+					Prose: strings.TrimSpace(criteria.Implementation),
+				})
```
xee5ch, line 91: I would recommend [per the OSCAL docs on `/catalog/parts/@name`](https://pages.nist.gov/OSCAL-Reference/models/v1.1.3/catalog/json-reference/#/catalog/groups/controls/parts) to change this to `example`.

```suggestion
				pts := append(*newCtl.Parts, oscal.Part{
					ID:    strings.TrimPrefix(criteria.ID, "OSPS-") + "_example",
					Name:  "example",
					Prose: strings.TrimSpace(criteria.Implementation),
				})
```

xee5ch (changes requested):

SecurityCRob (approved):

cmd/pkg/baseline/generator_oscal.go
puerco: Ah thanks for the tip! I've wrapped the doc into a "catalog" label and fixed the validation:

```
go run . validate --input-file ../ossf/security-baseline/cmd/oscal.json  
go-oscal: 2025/02/20 19:05:13 Successfully validated ../ossf/security-baseline/cmd/oscal.json is valid OSCAL version 1.1.3 catalog
```

cmd/pkg/baseline/generator_oscal.go
puerco: So this is the version string of the baseline the generator renders. This will be automatically picked up from the relases once we have them. For now it is hardcoded to"devel". I'll work on the tagging machinery once we cut our second release.

Thanks for the version-history @xee5ch! I'll make sure we generate those once we start releasing the oscal files as part of the releases .

cmd/pkg/baseline/generator_oscal.go
puerco: It is just a link back to the human-readable definition in the website, I updated the link refs to `canonical` as you suggested above.

cmd/pkg/baseline/generator_oscal.go
puerco: I took out the maturity level for now as I need to introduce some changes to the new schema to handle them properly.

cmd/pkg/baseline/generator_oscal.go
puerco: This field was dropped as part of the new schema definition.

cmd/pkg/baseline/generator_oscal.go
puerco: This field was dropped as part of the new schema

cmd/pkg/baseline/generator_oscal.go
puerco: This one was also deprecated with the new schema

puerco:
@xee5ch I see that the NIST namespace url (https://csrc.nist.gov/ns/oscal) does not return anything. Does this mean that the IRI is just used as a label?

Going through the docs it seems that the namespace does not really need to be defined anywhere, other linked data formats namespace with an URI/IRI but also expect a definition file (for example the Context in JSONLD).

I guess we are fine just defining our namespace and not really worrying about writing a definition?

xee5ch:
Correct it is just a label.

puerco:
OK, I pushed commits with updates for the new schema and to fix the validation with @defenseunicorns go-oscal.

The new output is published here: https://gist.github.com/puerco/e4a18354c281d45fc900cfb626adf286

cmd/pkg/baseline/generator_oscal.go
xee5ch: Sounds good, I thought you would appreciate that. From NIST or other community adopters, you are one of the few to take this tact and it is the smart play. Happy to see it!

cmd/pkg/baseline/generator_oscal.go
```diff
@@ -0,0 +1,120 @@
...
+				parts = append(parts, oscal.Part{
+					Class: control.ID,
+					ID:    ar.ID,
+					Name:  ar.ID,
+					Ns:    "",
+					Parts: &[]oscal.Part{
+						{
+							ID:    ar.ID + ".R",
+							Name:  "recomemendation",
+							Ns:    OpenSSFNS,
```
xee5ch, line 71: I am going to try and dig into this over the weekend, but in the short-term you may want to correct the spelling of the part name.

```suggestion
							Name:  "recommendation",
							Ns:    OpenSSFNS,
```

I would say a higher-level issue from using [the maintained `oscal-cli` community fork](https://github.com/metaschema-framework/oscal-cli/) is that there may be specific OSCAL rules that say a part should have a certain name if it is in the "default" core NIST OSCAL namespace. Weirdly, yours is not. You did it right. However, the error still persists in the tooling, so I will look into that; it might be a bug on that end so you can safely ignore it. :-)

xee5ch (changes requested):

cmd/pkg/baseline/generator_oscal.go
xee5ch: Actually it seems it is related to the NIST's definition in the OSCAL model, not the tooling. More to follow after https://github.com/usnistgov/OSCAL/issues/2112 is resolved and released in a new model version. You helped find an interesting edge case!

puerco: I noticed it too, it also seems that it is ignoring namespaced names.

funnelfiasco (approved):

PR #198, opened 2025-02-21
Add dependabot config

puerco:
The tooling and github actions in the repo are growing enough to warrant turning on dependabot.


eddie-knight (approved):
good call

PR #199, opened 2025-02-21
OSCAL validation action

puerco:
This PR adds a github action to verify the OSCAL output of the baseline compiler whenever a PR is opened.

This PR introduces a new file but requires #194 to run, here is an example run:
https://github.com/puerco/security-baseline/actions/runs/13448472115/job/37578580471

### DIFF
--- a/.github/workflows/oscal-check.yaml
+++ b/.github/workflows/oscal-check.yaml
@@ -18,7 +18,7 @@ jobs:
       - name: clone-baseline
         id: clone-baseline
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
-            with:
+        with:
                 persist-credentials: false
 
       - name: clone-go-oscal
--- a/.github/workflows/oscal-check.yaml
+++ b/.github/workflows/oscal-check.yaml
@@ -1,0 +1,41 @@
+# SPDX-FileCopyrightText: Copyright 2025 The Open Source Project Security Baseline Authors
+# SPDX-License-Identifier: Apache-2.0
+---
+name: Validate OSCAL Output
+
+on:
+  pull_request: {}
+  workflow_dispatch: {}
+
+permissions:
+  contents: read
+
+jobs:
+  verify:
+    runs-on: ubuntu-latest
+    steps:
+      - name: clone-baseline
+        id: clone-baseline
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+
+      - name: clone-go-oscal
+        id: clone-go-oscal
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+            repository: defenseunicorns/go-oscal
+            path: .build/go-oscal/
+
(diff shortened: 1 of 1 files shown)


xee5ch (commented):
Not to be labor an obvious point, but oscal-club (actually yours truly as fun else contributed) [has a packaged action that wraps the `oscal-cli` tooling](https://github.com/oscal-club/oscal-cli-action), be it the classic [NIST version](https://github.com/usnistgov/oscal-cli) and an updated fork from former NIST staff in [the metaschema-framework community-fork](https://github.com/metaschema-framework/oscal-cli/) that does some more bells and whistles than just JSON Schema validation, including SARIF output so you can overly the results on the files themselves in GitHub user interface.

I just thought I would put that out there if you want to consider it. If not, I completely understand, but it might help test multiple toolchains for conformance.

puerco:
Ah nice, I added it to also verify with it (see [this workflow run](https://github.com/puerco/security-baseline/actions/runs/13448868309/job/37579646289) here) and it actually caught more details.

It is not failing the workflow when validation fails but apparently, this is on purpose. I will leave it like non-failing like this while I fix the output to also validate with the oscal cli. Thanks for the suggestion.

.github/workflows/oscal-check.yaml
```diff
@@ -0,0 +1,54 @@
...
+  contents: read
+
+jobs:
+  verify:
+    name: Setup and Generate
+    runs-on: ubuntu-latest
+    steps:
+      - name: clone-baseline
+        id: clone-baseline
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
```
funnelfiasco, line 20:
```suggestion
        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
            with:
                persist-credentials: false
```

.github/workflows/oscal-check.yaml
```diff
@@ -0,0 +1,54 @@
...
+
+      - name: clone-go-oscal
+        id: clone-go-oscal
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+            repository: defenseunicorns/go-oscal
+            path: .build/go-oscal/
```
funnelfiasco, line 27:
```suggestion
            path: .build/go-oscal/
            persist-credentials: false
```

funnelfiasco (changes requested):
The `checkout` action doesn't drop credentials by default, which can result in a leak.

Otherwise, this looks good to me and to [zizmor](https://woodruffw.github.io/zizmor/).

SecurityCRob (approved):

.github/workflows/oscal-check.yaml
```diff
@@ -0,0 +1,57 @@
...
+
+jobs:
+  verify:
+    name: Setup and Generate
+    runs-on: ubuntu-latest
+    steps:
+      - name: clone-baseline
+        id: clone-baseline
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+            with:
```
funnelfiasco, line 21:
```suggestion
        with:
```

Whoops. Mis-indented my suggestion.

funnelfiasco (approved):

PR #201, opened 2025-02-21
Merged AC-03 and AC-04

eddie-knight:
I noticed that these two have a similar control purpose, so I've merged them into one control with two requirements.

### DIFF
--- a/baseline/OSPS-AC.yaml
+++ b/baseline/OSPS-AC.yaml
@@ -98,10 +98,10 @@ controls:
 
   - id: OSPS-AC-03
     title: |
-      Prevent unintentional direct commits against the primary branch
+      Prevent the primary branch from unintentional or unauthorized changes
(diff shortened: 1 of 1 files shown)


baseline/OSPS-AC.yaml
```diff
@@ -98,10 +98,10 @@ controls:
 
   - id: OSPS-AC-03
     title: |
-      Prevent unintentional direct commits against the primary branch
+      Prevent the primary branch from unintentional or unauthorized changes
```
eddie-knight, line 101:
```suggestion
      Protect the primary branch from unintentional or unauthorized changes
```

funnelfiasco (approved):
This makes sense to me. I don't see any reason to treat certain modifications differently.

SecurityCRob (approved):

PR #205, opened 2025-02-22
Revising Vuln Management controls based on feedback

eddie-knight:
- merged SCA requirements into one control
- added requirements so that SAST and SCA are handled similarly
- renumbered the resulting SCA control to the bottom, a stylistic pre-release choice since all of these requirements are level
- added an explicit requirement for VEX at level 3
- also I'm pretty sure the impacted controls mappings are irreparably jankified, I'm sorry @SecurityCRob

### DIFF
--- a/baseline/OSPS-VM.yaml
+++ b/baseline/OSPS-VM.yaml
@@ -156,6 +156,7 @@ controls:
           The project MUST publicly publish data about discovered 
           vulnerabilities.
         applicability:
+          - Maturity Level 2
           - Maturity Level 3
         recommendation: |
           Provide information about known vulnerabilities in a predictable
@@ -163,11 +164,22 @@ controls:
           To the degree possible, this information should include affected
           version(s), how a consumer can determine if they are vulnerable, and
           instructions for mitigation or remediation.
+      - id: OSPS-VM-04.02
+        text: |
+          The project's released software assets MUST include VEX data to
+          provide information about the exploitability of vulnerabilities.
+        applicability:
+          - Maturity Level 3
+        recommendation: |
+          Include a VEX file or VEX data in an SBOM as part of the project's 
+          released software assets. This data should provide information about
+          the exploitability of vulnerabilities, including which conditions or
+          configurations are necessary for the vulnerability to be exploited
+          or non-applicable.
 
   - id: OSPS-VM-05
     title: |
-      Define and enforce a threshold for remediation of SCA findings related to
-      vulnerabilities and licenses
+      Define and enforce a threshold for remediation of SCA findings
     objective: |
       Ensure that the project clearly communicates the threshold for remediation
       of SCA findings, including vulnerabilities and license issues in software
--- a/baseline/OSPS-VM.yaml
+++ b/baseline/OSPS-VM.yaml
@@ -261,7 +261,7 @@ controls:
         text: |
           All changes to the project's codebase MUST be automatically evaluated
           against a documented policy for malicious dependencies and
-          known vulnerabilities in depenencies and blocked in the event of
+          known vulnerabilities in dependencies and blocked in the event of
           violations except when declared and suppressed as non-exploitable.
         applicability:
           - Maturity Level 3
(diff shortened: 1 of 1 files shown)


baseline/OSPS-VM.yaml
```diff
@@ -284,16 +151,109 @@ controls:
...
+          evaluated against a documented policy for known vulnerabilities and
+          blocked in the event of violations except when declared and suppressed
+          as non-exploitable.
+        applicability:
+          - Maturity Level 3
+        recommendation: |
+          Create a status check in the project's version control system that
+          runs a Static Application Security Testing (SAST) tool on all changes
+          to the codebase. Require that the status check passes before changes
+          can be merged.
```
xee5ch, line 259: The related policy above for OSPS-VM-05.03 requires a documented policy for known vulnerabilities but this recommendation described SAST testing to identify known vulnerabilities. I would have presumed SCA testing for dependency analysis, but I usually do not have industry colleagues describe SAST and SCA equivalent. Is this the intent?

eddie-knight: I think you caught me in an oversight 🤗

baseline/OSPS-VM.yaml
```diff
@@ -284,16 +151,145 @@ controls:
...
           Provide information about known vulnerabilities in a predictable
           public channel, such as a CVE entry, blog post, or other medium.
           To the degree possible, this information should include affected
           version(s), how a consumer can determine if they are vulnerable, and
           instructions for mitigation or remediation.
+
+  - id: OSPS-VM-05
+    title: |
+      Define and enforce a threshold for remediation of SCA findings related to
+      vulnerabilities and licenses
```
eddie-knight, line 170:
```suggestion
      Define and enforce a threshold for remediation of SCA findings
```

TheFoxAtWork (approved):
1 spelling correction.  LGTM

baseline/OSPS-VM.yaml
```diff
@@ -297,3 +164,144 @@ controls:
           To the degree possible, this information should include affected
           version(s), how a consumer can determine if they are vulnerable, and
           instructions for mitigation or remediation.
+      - id: OSPS-VM-04.02
+        text: |
+          The project's released software assets MUST include VEX data to
+          provide information about the exploitability of vulnerabilities.
```
puerco, line 170: Two notes here:
1. This should be a MUST only when dealing with false positives. 

2. You can include a snapshot of the applicable data in the assets, but the VEX feed keeps moving after the release, so we prefer to have a continuously updating feed.

```suggestion
          Any vulnerabilities in the software components not affecting the
          project MUST be accounted for in a VEX document, augmenting
          the vulnerability report with non-exploitability details.
```

baseline/OSPS-VM.yaml
```diff
@@ -297,3 +164,144 @@ controls:
...
+        applicability:
+          - Maturity Level 3
+        recommendation: |
+          Include a VEX file or VEX data in an SBOM as part of the project's 
+          released software assets. This data should provide information about
+          the exploitability of vulnerabilities, including which conditions or
+          configurations are necessary for the vulnerability to be exploited
+          or non-applicable.
```
puerco, line 178: Discourage VEX in SBOMs. While it is perfectly acceptable to use the SBOM formats to communicate VEX, we want to keep the VEX data separate from the bill of materials

```suggestion
          Establish a VEX feed communicating the exploitability status of 
          known vulnerabilities, including assessment details or any
          mitigations in place preventing vulnerable code from being
          executed.
```

puerco (commented):
The goal with vex should be to provide a stream of machine-readable data of non-exploitability while, at the same time, discouraging human content (PDFs, blog posts) as the only messaging channel for false positive data. But we should only require it when it needs to exist.

I suggested some changes below, feel free to english them good :point_down:

baseline/OSPS-VM.yaml
eddie-knight: thanks for this! Working to refine it... Where should the VEX document live? in any part of the project documentation, or specifically alongside the corresponding release assets?

SecurityCRob (approved):

PR #211, opened 2025-02-24
Fix homophone error in README.md

semioticrobotic:
Replace "it's" with "its"


puerco (approved):

funnelfiasco (approved):
Good catch @semioticrobotic (and good to see you here)

PR #216, opened 2025-02-25
Refactor Legal Requirements & Titles

eddie-knight:


baseline/OSPS-LE.yaml
```diff
@@ -35,9 +36,9 @@ controls:
     assessment-requirements:
       - id: OSPS-LE-01.01
         text: |
-          The version control system MUST require all code contributors to assert
-          that they are legally authorized to commit the associated contributions
-          on every commit.
+          While active, the version control system MUST require all code
```
funnelfiasco, line 39: Does "while active" make sense here?

baseline/OSPS-LE.yaml
```diff
@@ -73,8 +76,8 @@ controls:
     assessment-requirements:
       - id: OSPS-LE-02.01
         text: |
-          The license for the source code MUST meet the OSI Open Source Definition
-          or the FSF Free Software Definition.
+          While active, the license for the source code MUST meet the OSI Open
```
funnelfiasco, line 79: Ditto

SecurityCRob (approved):

funnelfiasco (approved):
I'm not sure the "while active"s make sense here (I stopped commenting after two), but I won't block on it. They're not harmful, just maybe meaningless?

baseline/OSPS-LE.yaml
eddie-knight: Couldn't think of another "while" statement... got any suggestions?

eddie-knight: Also I couldn't convince myself to break pattern after doing this like 20 times before :D

funnelfiasco: I'm not sure it needs a while in this case. It's basically `while(True)`. I guess if we want to be sticklers for the format, it's fine, but like...if the project is inactive, it doesn't matter 🤷 

F it. Let's leave it as is and if it causes confusion we can fix it later.

PR #224, opened 2025-02-25
Created metadata.yaml for reference to reusable elements

eddie-knight:


funnelfiasco:
While this is still in draft status, I'd like to suggest using "LevelX" instead of "MX" for levels. It's more characters, but it also makes it more self-evident what it means. I'd accept "LX", but I'd be sad about it.

baseline/metadata.yaml
```diff
@@ -0,0 +1,50 @@
+metadata:
+  id: OSPS-B
+  title: Open Source Project Security Baseline
+  description: |
+    The Open Source Project Security (OSPS) Baseline is a set of security
+    criteria that projects should meet to demonstrate a strong security posture.
+  applicability-categories:
+    - id: Maturity Level 1
```
puerco, line 8: Constrain the identifiers to strings without spaces. These will get composed and fed into other systems that could may not handle multiple words
```suggestion
    - id: MA-1
```

SecurityCRob (approved):
saving money by removing characters.... a bold strategy Cotton!

SecurityCRob:
> saving money by removing characters.... a bold strategy Cotton!

oh SNAP!  unintentional name collision with well-known meme.  Double-points to Ravenclaw!

eddie-knight:
I'll reopen this later with a larger and more complete proposal

PR #227, opened 2025-02-26
Update lexicon.yaml with Scorecard entry

SecurityCRob:
added out pals at Scorecard to our control mapping

### DIFF
--- a/baseline/lexicon.yaml
+++ b/baseline/lexicon.yaml
@@ -218,7 +218,6 @@
      to help improve the health of critical projects
      that the community depends on.
   synonyms: 
-    - ScCrd 
   references: 
     - https://github.com/ossf/scorecard
     - https://scorecard.dev/
--- a/baseline/lexicon.yaml
+++ b/baseline/lexicon.yaml
@@ -214,7 +214,7 @@
   definition: |
      An OpenSSF project that helps users assesses open 
      source projects for security risks through a series 
-     of automated checksIt was created by OSS developers 
+     of automated checks. It was created by OSS developers 
      to help improve the health of critical projects
      that the community depends on.
   synonyms: 
--- a/baseline/lexicon.yaml
+++ b/baseline/lexicon.yaml
@@ -210,6 +210,18 @@
   references: 
     - https://www.opencre.org/
     - https://zeljkoobrenovic.github.io/opencre-explorer/ 
+- term: OpenSSF Scorecard
+  definition: |
+     An OpenSSF project that helps users assesses open 
+     source projects for security risks through a series 
+     of automated checksIt was created by OSS developers 
+     to help improve the health of critical projects
+     that the community depends on.
+  synonyms: 
+    - ScCrd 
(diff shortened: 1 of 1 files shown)


baseline/lexicon.yaml
```diff
@@ -210,6 +210,18 @@
   references: 
     - https://www.opencre.org/
     - https://zeljkoobrenovic.github.io/opencre-explorer/ 
+- term: OpenSSF Scorecard
+  definition: |
+     An OpenSSF project that helps users assesses open 
+     source projects for security risks through a series 
+     of automated checksIt was created by OSS developers
```
funnelfiasco, line 217:
```suggestion
     of automated checks. It was created by OSS developers 
```

baseline/lexicon.yaml
```diff
@@ -210,6 +210,18 @@
...
     - https://zeljkoobrenovic.github.io/opencre-explorer/ 
+- term: OpenSSF Scorecard
+  definition: |
+     An OpenSSF project that helps users assesses open 
+     source projects for security risks through a series 
+     of automated checksIt was created by OSS developers 
+     to help improve the health of critical projects
+     that the community depends on.
+  synonyms: 
+    - ScCrd
```
funnelfiasco, line 221: Really?

funnelfiasco (approved):

baseline/lexicon.yaml
SecurityCRob: suggestions welcome

eddie-knight: I think Ben means to ask "Why?"

Synonyms aren't required, just used to apply links elsewhere in the document if that synonym is found.

eddie-knight (approved):

baseline/lexicon.yaml
eddie-knight: Ah, I see PR #228 has a framework mapping to `ScCrd`. This is the wrong-ish place for that — having a lexicon entry is fine, but mappings should point to the framework list that we created.

While working on #224 I was thinking about the overlap between the frameworks and the lexicon. I think I can help to make this work how you intended... I'll ping you when I have it ready.

eddie-knight: Okay, no large changes... I think we should allow duplication between the mappings and lexicon, so that when folks click on "Scorecard" they are brought here for a long-form description... And when folks look for a list of mappings, we can send them to the shortlist without the extras.

The alternative would require the mapping references to have description AND synonyms as a field, which... just didn't feel right.

I added scorecard as a mapping in #224, so we don't need the synonym here.

```suggestion
```

SecurityCRob: so you pulled all the things that were formerly in Lexicon and moved them to Frameworks?  Am I understanding that correctly?

eddie-knight: didn't move anything — right now it's duplicated as I haven't come up with a design to avoid that duplication

PR #228, opened 2025-02-26
Update OSPS-AC.yaml with mapping corrections/updates

SecurityCRob:
latest pass at validating compliance crosswalk mappings for AC group

### DIFF
--- a/baseline/OSPS-AC.yaml
+++ b/baseline/OSPS-AC.yaml
@@ -30,9 +30,11 @@ controls:
         identifiers:
           - PO3.2
           - PS1
+          - PS2
       - reference-id: CSF
         identifiers:
           - PR.A-02
+          - PR.A-05
       - reference-id: OCRE
         identifiers:
           - 486-813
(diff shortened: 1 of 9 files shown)


baseline/OSPS-AC.yaml
```diff
@@ -183,6 +193,10 @@ controls:
           - 347-507
           - 263-284
           - 123-124
+        - reference-id: SLSA
+          identifiers:
+          - Producer - Choose an appropriate build platform
+          - Build platform - Isolation strength - Isolated
```
funnelfiasco, line 199:
```suggestion
      - reference-id: SLSA
        identifiers:
          - Producer - Choose an appropriate build platform
          - Build platform - Isolation strength - Isolated
```

I lied. This fixes it. I should have tried locally first :-(

funnelfiasco (approved):
I trust your judgment on the mappings. But it compiles now!

I'm not sure why the DCO check is failing, but whatever. You can squash my commits into your branch and then it will all be good.

eddie-knight (approved):

PR #244, opened 2025-03-11
normalize NIST SSDF references to XX.0.0 format

maennchen:
NIST SSDF uses the format `XX.0.0` (including a dot after the Section before the Practice.

Currently there's a mix of `.`, nothing and a whitespace. This PR normalizes it to use a dot as present in the NIST SSDF pdf.

image


eddie-knight (approved):

eddie-knight:
Thanks for finding and fixing these!

funnelfiasco (approved):
Good catch!

Issue #252, opened 2025-03-27
Propose changing DO-04 and DO-05 recommendation language


trumant:
While the baseline control recommendations suggest that support statements are expected to appear in either `SUPPORT.md` or `README.md` under a heading of Support, the two projects I could quickly find that actually document their support policy do so in `.github/SECURITY.md`

- https://github.com/privateerproj/privateer-sdk/security/policy
- https://github.com/kubernetes/kubernetes/security/policy

The recommendations also talk about an OpenEoX file, but I cannot find any indication that oasis has achieved any momentum on that beyond oasis-tcs/openeox#15

## Proposed changes

- [x] Remove mentions of OpenEoX until such time as momentum from Oasis or others in the community indicate meaningful utility and adoption
- [x] Update recommendation language to match documentation patterns more broadly observed within the FLOSS community
- [ ] Consider and potentially propose updates to the control and requirement language of DO-04 as it fails to provide clear examples of relevant support statements such as "this project is abandoned" or "seeking maintainers, support limited"

trumant:
Related conversation around OpenEoX: https://github.com/oasis-tcs/openeox/issues/64

eddie-knight:
I agree with the assessment here. Thanks for starting the oasis conversation @evankanderson and thanks for the suggestion/PR here @trumant!

trumant:
> [@trumant](https://github.com/trumant) [#253](https://github.com/ossf/security-baseline/pull/253) seems to address your first two proposed changes. Do you want to propose an edit for your third change, or should we close this issue?
> 
> (FWIW, I'm not sure how far down the defining support levels rabbit hole we'd want to go. I had a thought the other day that we should have an RFC that explicitly defines these things, but I absolutely do not want to do that myself. Actually, I do, but I know better)

@funnelfiasco I think we can close this issue and possibly carry over that third unresolved item into a new, separate issue if we feel it may be needed down the road.

PR #253, opened 2025-03-27
Removing DO-04 and DO-05 recommendations for OpenEoX usage

trumant:
This PR partially addresses issue #252 by removing recommendations around OpenEoX and adding a bit more "color" in the recommendations

### DIFF
--- a/baseline/OSPS-DO.yaml
+++ b/baseline/OSPS-DO.yaml
@@ -202,7 +202,8 @@ controls:
         recommendation: |
           In order to communicate the scope and duration of support for the
           project's released software assets, the project should have a
-          SUPPORT.md or other documentation explaining the support lifecycle,
+          SUPPORT.md file, a "Support" section in SECURITY.md, or
+          other documentation explaining the support lifecycle,
           including the expected duration of support for each release, the
           types of support provided (e.g., bug fixes, security updates), and
           any relevant policies or procedures for obtaining support.
--- a/baseline/OSPS-DO.yaml
+++ b/baseline/OSPS-DO.yaml
@@ -5,7 +5,8 @@ description: |
   of the project. These controls help ensure that
   the project's documentation is comprehensive,
   accurate, and up-to-date, enabling users to
-  understand the project's features and functionality.
+  understand the project's features and functionality, maintenance, support,
+  security and release practices.
 controls:
   - id: OSPS-DO-01
     title: |
@@ -201,7 +202,10 @@ controls:
         recommendation: |
           In order to communicate the scope and duration of support for the
           project's released software assets, the project should have a
-          SUPPORT.md or an OpenEoX file in a well known location.
+          SUPPORT.md or other documentation explaining the support lifecycle,
+          including the expected duration of support for each release, the
+          types of support provided (e.g., bug fixes, security updates), and
+          any relevant policies or procedures for obtaining support.
 
   - id: OSPS-DO-05
     title: |
@@ -234,9 +238,9 @@ controls:
         applicability:
           - Maturity Level 3
         recommendation: |
-          While a machine-readable OpenEoX file is recommended, this may also be
-          communicated in a SUPPORT.md or beneath a Support header in the
-          primary README.md.
+          In order to communicate the scope and duration of support for security
+          fixes, the project should have a SUPPORT.md or other documentation
+          explaining the project's policy for security updates.
 
   - id: OSPS-DO-06
     title: |


eddie-knight (approved):

baseline/OSPS-DO.yaml
```diff
@@ -201,7 +202,10 @@ controls:
         recommendation: |
           In order to communicate the scope and duration of support for the
           project's released software assets, the project should have a
-          SUPPORT.md or an OpenEoX file in a well known location.
+          SUPPORT.md or other documentation explaining the support lifecycle,
```
evankanderson, line 205: Did you want to include the suggestion about using `SECURITY.md` that you had in #252 ?

```suggestion
          SUPPORT.md file, a "Support" section in SECURITY.md, or
          other documentation explaining the support lifecycle,
```

baseline/OSPS-DO.yaml
```diff
@@ -234,9 +238,9 @@ controls:
         applicability:
           - Maturity Level 3
         recommendation: |
-          While a machine-readable OpenEoX file is recommended, this may also be
-          communicated in a SUPPORT.md or beneath a Support header in the
-          primary README.md.
+          In order to communicate the scope and duration of support for security
+          fixes, the project should have a SUPPORT.md or other documentation
+          explaining the project's policy for security updates.
```
evankanderson, line 243: I'd love for level 3 to include a machine-readable requirement; possibly something along the lines of:

```suggestion
          In order to communicate the scope and duration of support for security
          fixes, the project must have a machine-readable version support matrix.
          This support matrix should be readable by common aggregation software,
          such as https://endoflife.date/.
```

(I'd love to have a stronger standard to point to, but it feels like level 3 should really support machine-processing of "is my version supported")

evankanderson (commented):
My $0.02, though I understand others may have different yardsticks.

baseline/OSPS-DO.yaml
eddie-knight: brainstorming, but... we might also want to add this into the project section of security insights.

https://github.com/ossf/security-insights-spec/blob/main/template-full.yml

trumant: Yep, agreed. I will open a security insights enhancement proposal

trumant: https://github.com/ossf/security-insights-spec/issues/112

trumant: While I agree strongly in the value of machine-readable EoL/support data, I think it makes sense to start with the insights spec change I've proposed above which at least gets us to the point where maintainers can define where this content is documented for human user consumption.

A further insights schema update can also be proposed that introduces a URL to the machine-readable form of that same data. WDYT?

evankanderson: I'm slightly concerned about having some parts of security-insights be fast-moving (i.e. needing updates on each release) while other parts are not.  In my experience, it becomes harder to automate _partial_ changes to a file compared with being able to regenerate the whole file (in this case, a version support table).

That's not to say that it can't go into Security Insights, just that I would probably put release-specific information in a separate file from documentation, maintainer, etc data.

evankanderson: To state my position on this PR clearly, I'm okay with starting here, but my personal opinion is that if the support documentation isn't machine-consumable, we've missed the real value for _consumers_ of open source.  Ideally, for level 2 or level 3 projects, they should be able to automatically flag when they are producing or deploying an artifact which is outside the upstream support window, and decide how they want to mitigate that risk.

trumant: > for level 2 or level 3 projects, they should be able to automatically flag when they are producing or deploying an artifact which is outside the upstream support window, and decide how they want to mitigate that risk

This is interesting - I could imagine this type of control becoming `OSPS-QA-02.03` perhaps

baseline/OSPS-DO.yaml
```diff
@@ -201,7 +202,11 @@ controls:
         recommendation: |
           In order to communicate the scope and duration of support for the
           project's released software assets, the project should have a
-          SUPPORT.md or an OpenEoX file in a well known location.
+          SUPPORT.md file, a "Support" section in SECURITY.md, or
```
trumant, line 205: Given the diversity of existing practices across the community illustrated in https://github.com/ossf/security-insights-spec/pull/111 we may want to reconsider this prescriptive guidance about `SUPPORT.md` and `SECURITY.md`

If the changes in that PR merge, should we consider updating this to be:

```diff
-          project's released software assets, the project should have a
-          SUPPORT.md file, a "Support" section in SECURITY.md, or
-          other documentation explaining the support lifecycle,
-          including the expected duration of support for each release, the
-          types of support provided (e.g., bug fixes, security updates), and
-          any relevant policies or procedures for obtaining support.
+          project's released software assets, the project publishes and
+          maintains documentation explaining the support lifecycle, including
+          the expected duration of support for each release, the types of
+          support provided (e.g., bug fixes, security updates), and any relevant
+          policies or procedures for obtaining support.
```
and perhaps likewise for the change below, the update becomes:

```diff
           - Maturity Level 3
         recommendation: |
           In order to communicate the scope and duration of support for security
-          fixes, the project should have a SUPPORT.md or other documentation
-          explaining the project's policy for security updates.
+          fixes, the project publishes and maintains documentation explaining
+          the project's policies for security updates.
```

evankanderson: It seems like endoflife.date is able to parse webpages with this info, so maybe that is a reasonable standard to push people towards

funnelfiasco: There's value in providing explicit examples here as it gives some useful guidance to project maintainers. The less we make them have to figure out on their own, the more useful Baseline will be (and the better our adoption will be). So if the security-insights-spec PR is merged, I'd be fine with making that our recommendation instead of/in addition to the SUPPORT.md  and SECURITY.md references, but we shouldn't drop it entirely.

funnelfiasco (approved):
Thanks @trumant!

PR #259, opened 2025-04-08
Add a control for unreviewable binaries

### LINKED ISSUE #242: Baseline references "generated executables", but not "unreviewable binaries"
xz-utils attack included files which weren't introspectable/reviewable which allowed code to be smuggled into the source repo. Maybe [OSPS-QA-05.01](https://baseline.openssf.org/versions/2025-02-25-rc#osps-qa-0501) can have this criteria be added?

funnelfiasco:
This compliments OSPS-QA-05.01.

I chose to make it separate from OSPS-QA-05.01 instead of modifying the language because they seem like slightly different issues. 

Fixes #242

### DIFF
--- a/baseline/OSPS-QA.yaml
+++ b/baseline/OSPS-QA.yaml
@@ -290,6 +290,20 @@ controls:
           executable artifact appears critical to a process such as testing, it
           should be instead be generated at build time or stored separately and
           fetched during a specific well-documented pipeline step.
+      - id: OSPS-QA-05.02
+        text: |
+          While active, the version control system MUST NOT contain unreviewable
+          binary artifacts.
+        applicability:
+          - Maturity Level 1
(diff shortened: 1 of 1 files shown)


eddie-knight (approved):
FWIW, there are some exceptions that Scorecard tolerates which we do not capture in this language

funnelfiasco:
> FWIW, there are some exceptions that Scorecard tolerates which we do not capture in this language

Should we?

eddie-knight:
Chatted with @hboutemy about the gradle/maven wrapper exceptions, and I think we're safe to move without any carveouts or exceptions.

We should just be aware that some folks will feel like this is debatable due to a long-standing risk tolerance for "verified" binaries being stored in the repo.

SecurityCRob (approved):

PR #280, opened 2025-04-30
Use stdout where appropriate

### LINKED ISSUE #279: Use stdout instead of stderr for non-error messages
The go tooling uses `os.Stderr` in a lot of places where the output is informational. We should use `os.Stdout` instead. This better conforms to the way CLI applications Should Behave™.

funnelfiasco:
Fixes #279

### DIFF
--- a/cmd/internal/cmd/compile.go
+++ b/cmd/internal/cmd/compile.go
@@ -117,14 +117,14 @@ func addCompile(parentCmd *cobra.Command) {
 				if err := gen.ExportMarkdown(bline, opts.templatePath, opts.outPath); err != nil {
 					return fmt.Errorf("writing markdown render: %w", err)
 				}
-				fmt.Fprintf(os.Stderr, "\n✅ Baseline rendered to %s\n", opts.outPath)
+				fmt.Fprintf(os.Stdout, "\n✅ Baseline rendered to %s\n", opts.outPath)
 			}
 
-			fmt.Fprintf(os.Stderr, "\nℹ️  Counts\n")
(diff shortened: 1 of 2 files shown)


eddie-knight (approved):

cmd/internal/cmd/compile.go
```diff
@@ -117,22 +117,22 @@ func addCompile(parentCmd *cobra.Command) {
 				if err := gen.ExportMarkdown(bline, opts.templatePath, opts.outPath); err != nil {
 					return fmt.Errorf("writing markdown render: %w", err)
 				}
-				fmt.Fprintf(os.Stderr, "\n✅ Baseline rendered to %s\n", opts.outPath)
+				fmt.Fprintf(os.Stdout, "\n✅ Baseline rendered to %s\n", opts.outPath)
```
puerco, line 120:
```suggestion
				fmt.Printf("\n✅ Baseline rendered to %s\n", opts.outPath)
```

puerco (commented):
Thanks for changing these @funnelfiasco! Just one nit: change these fmt.Fprintf to STDOUT to regular printfs. This is redundancy that the linter should catch, I'll set it up in a follow -up PR.

SecurityCRob (approved):

puerco:
OK, so I meant correcting all the instances, not just the one suggestion. I guess we can fix it later.

funnelfiasco:
> OK, so I meant correcting all the instances, not just the one suggestion. I guess we can fix it later.

My mistake! I'll make that change now in a follow-up

PR #283, opened 2025-04-30
Document the baseline-compiler

### LINKED ISSUE #278: Document the use of the generation tool
We don't actually say how to use the go tooling in `cmd/`, which seems like a thing we should do.

funnelfiasco:
Not the world's greatest documentation, but it's better than what we have. 😄 

Fixes #278

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ From the `cmd/` directory, run `go run . [command] [arguments]`
 
 ### compile
 
-This command reads the yaml contents and generates two optional Markdown files: a listing of all OSPS Baseline controls for use in a static site generator and a checklist of controls by level.
+This command reads the YAML data describing the Baseline controls in `/baseline` and generates two optional Markdown files: a listing of all OSPS Baseline controls for use in a static site generator and a checklist of controls by level.
 
 To produce a file for use in a static site generator (which is how we create baseline.openssf.org!), specify a file argument to `--output`.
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -38,6 +38,38 @@ Each entry has the following values:
...
+
+To produce a file for use in a static site generator (which is how we create baseline.openssf.org!), specify a file argument to `--output`.
+
+To produce a checklist file to use as an aid to evaluating your project's conformance to OSPS Baseline, specify a file argument to `--checklist-output`.
+
+Use the `--help` flag for more options.
+
+### oscal
+
+This command validates the OSPS Baseline output against [OSCAL](https://pages.nist.gov/OSCAL/). By default, it writes the JSON output to STDOUT. You can specify a file with `--out`.
```
trumant, line 65: I think it converts the catalog controls in the baseline from a SCI Layer 2 YAML format to an OSCAL JSON format. I read the code and I don't see any explicit validation of the OSCAL payload against any JSON schema, but that me be happening within the oscal module used.

@eddie-knight would know better, I suspect

trumant (changes requested):

README.md
puerco: We have a presubmit that checks the oscal output:

https://github.com/ossf/security-baseline/actions/runs/14767602657/job/41461983532

trumant (approved):

puerco (approved):

PR #284, opened 2025-05-01
feat: use the SCI layer2 module for the cmd CLI

trumant:
This PR removes duplication of the layer2 code
from revanite-io/sci in favor of simply using that module directly.

This PR also changes our YAML dependency as the prior one used is now archived and no longer maintained.


trumant:
@eddie-knight and @jmeridth FYI that while this mostly "works", it also demonstrates some continued incompatibilities between SCI and the way the catalog is maintained in this repository.

Here are some key problems in the the Markdown produced by `cd cmd && go run ./... compile --output foo.md` and the reason why the problem exists:

## Applicability is broken

```markdown
## Controls Overview

* [Level 1](#level-1): for any code or non-code project with any number of maintainers or users
* [Level 2](#level-2): for any code project that has at least 2 maintainers and a small number of consistent users
* [Level 3](#level-3): for any code project that has a large number of consistent users

### Level 1


### Level 2


### Level 3
```

This issue exists because https://pkg.go.dev/github.com/revanite-io/sci@v0.3.4/pkg/layer2#Control expects:

```yaml
id: OSPS-AC-01
    title: |
      The project's version control system MUST require multi-factor
      authentication for collaborators modifying the project repository
      settings or accessing sensitive data.
    objective: |
      Reduce the risk of account compromise or insider threats by requiring
      multi-factor authentication for collaborators modifying the project
      repository settings or accessing sensitive data.
    mappings:
      - reference-id: BPB
        identifiers:
          - CC-G-1
      - reference-id: CRA
        identifiers:
          - 1.2d
          - 1.2e
          - 1.2f
      - reference-id: SSDF
        identifiers:
          - PO.3.2
          - PS.1
          - PS.2
      - reference-id: CSF
        identifiers:
          - PR.A-02
          - PR.A-05
      - reference-id: OCRE
        identifiers:
          - 486-813
          - 124-564
          - 347-352
          - 333-858
          - 152-725
          - 201-246
    assessment-requirements:
      - id: OSPS-AC-01.01
        text: |
          When a user attempts to access a sensitive resource in the project's
          version control system, the system MUST require the user to complete
          a multi-factor authentication process.
        applicability:
          - Maturity Level 1
          - Maturity Level 2
          - Maturity Level 3
        recommendation: |
          Enforce multi-factor authentication for the project's version
          control system, requiring collaborators to provide a second form of
          authentication when accessing sensitive data or modifying repository
          settings. Passkeys are acceptable for this control.
```
specify a field for Control named `assessment-requirements` but SCI expects `requirements`.

This, in turn, prevents the template logic from ever having applicability data in the requirements to use to group by level.

There may be other Layer 2 issues to sort out, but that's what I've found so far.

trumant:
The `requirements` vs `assessment-requirements` issue is also the root cause of the failure in the OSCAL workflow seen in https://github.com/ossf/security-baseline/actions/runs/14775559544/job/41483075516?pr=284

SecurityCRob (dismissed):

SecurityCRob:
Are we interested in making the change from assessment-requirements to requirements?  I'm glad to go through files and make changes if the group agrees.

trumant:
> Are we interested in making the change from assessment-requirements to requirements? I'm glad to go through files and make changes if the group agrees.

In this particular case, I believe the change should probably happen in SCI, because of:
- https://github.com/revanite-io/sci/blob/v0.3.4/schemas/layer-2.cue#L44he S
- https://github.com/revanite-io/sci/blob/v0.3.4/pkg/layer2/types.go#L40

From my perspective, this is more motivation to ensure that the go types for Layer 2 are generated based on the schema, so they cannot fall out of sync with one another.

That said, since the baseline YAML files do not contain a schema reference to the version of the SCI layer 2 schema in use, its hard to pin down.

In short, this is good feedback for the SCI project and can help us mature it

funnelfiasco (changes requested):
Thanks for making this better, @trumant! Nack-ing for now so that we don't accidentally merge this too soon.

I think the overall approach is good, but it will also require an update to template-checklist.md and to README.md to reflect the changed name of the fields. Plus the open discussion items.

trumant:
> Are we interested in making the change from assessment-requirements to requirements? I'm glad to go through files and make changes if the group agrees.

I've made these changes within this PR

trumant:
> Thanks for making this better, @trumant! Nack-ing for now so that we don't accidentally merge this too soon.
> 
> I think the overall approach is good, but it will also require an update to template-checklist.md and to README.md to reflect the changed name of the fields. Plus the open discussion items.

Thanks for the reminder on those additional required changes. They have now been made.

funnelfiasco (dismissed):
Looks good to me! In addition to the CI checks, I ran it manually and published it to my fork and everything seems good. And it's a net reduction of 51 lines! 🎉

cmd/internal/cmd/compile.go
```diff
@@ -121,8 +121,8 @@ func addCompile(parentCmd *cobra.Command) {
 			}
 
 			fmt.Printf("\nℹ️  Counts\n")
-			for c := range bline.Categories {
-				fmt.Printf(" OSPS-%s: %d controls\n", c, len(bline.Categories[c].Controls))
+			for _, family := range bline.Catalog.ControlFamilies {
+				fmt.Fprintf(os.Stderr, " OSPS-%s: %d controls\n", bline.ControlFamilyIDs[family.Title], len(family.Controls))
```
eddie-knight, line 125: Any reason this is changed to an error out? If I recall this section correctly, it was just intended to give some quick feedback during manual compilation.

eddie-knight (approved):
This is excellent @trumant 🔥 

And @funnelfiasco thanks for testing locally so that I can just review the code changes here ✊

eddie-knight (approved):

SecurityCRob (commented):
+1

PR #290, opened 2025-05-05
Split spell checking into a separate workflow

funnelfiasco:
Run the spellcheck separately from the build so that it makes CI failures more immediately clear (and gives us the option to make build failures blocking but have spelling failures be not-blocking later if we so choose. I am not advocating that, just noting that it gives us the option)

Signed-off-by: Ben Cotton <ben@kusari.dev>

### DIFF
--- a/.github/workflows/spellcheck.yml
+++ b/.github/workflows/spellcheck.yml
@@ -1,0 +1,21 @@
+name: Spell check
+
+on:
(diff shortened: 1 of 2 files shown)


SecurityCRob (approved):

puerco (approved):
Thanks!

Issue #281, opened 2025-04-30
`oscal` and `verify` commands do not have a default argument to `-b`


funnelfiasco:
The compile command does. I think it's reasonable to not make people type `-b ../baseline` every time they want to run the check.

puerco:
I think I prefer running all commands from the repo root, but since we have one, I opened https://github.com/ossf/security-baseline/pull/294 to even them out.

PR #294, opened 2025-05-05
Even baseline path in all commands

puerco:
This PRs introduces a default path constant with the path to look for the baseline YAML files. It is now shared across all subcommands.

Closes https://github.com/ossf/security-baseline/issues/281

Signed-off-by: Adolfo García Veytia (Puerco) <puerco@carabiner.dev>


SecurityCRob (approved):

eddie-knight (approved):
nice ncie, LGTM

Issue #275, opened 2025-04-29
Should we explicitly contemplate AI actors in Baseline?


funnelfiasco:
While discussing #255, Daniel Thompson-Yvetot said "we will soon start seeing virtually autonomous “AI” contributions in whole and in part, maybe this deserves consideration as well…"

Right now the Baseline does not explicitly contemplate AI actors, although it did come up some in the discussion of PR reviews. It's worth considering how we might want to handle this, if at all.

Note that I am only talking about AI actors operating on a repo, CI/CD system, etc. Requirements for producing AI models is out of scope for this particular discussion.

trumant:
Agreed it deserves some consideration. My take is that contributions like this are vaguely addressed by https://baseline.openssf.org/versions/2025-02-25#osps-gv-0302 and https://baseline.openssf.org/versions/2025-02-25#osps-gv-0401 and https://baseline.openssf.org/versions/2025-02-25#osps-qa-0701 today

Perhaps we should consider adding AI contributor or AI-assisted contributor recommendations to those.

eddie-knight:
Is this covered by the usage of the term "entity" in the definition of a contributor?

> Entities who commit code or documentation to the project. Code contributors include collaborators or external participants who submit changes.

funnelfiasco:
> Is this covered by the usage of the term "entity" in the definition of a contributor?

Implicitly? I'd say so. The questions are:

1. Do we want to explicitly refer to AI actors?
2. Should there be any special-case considerations for AI actors?

trumant:
> maintainer and contributor should be explicitly human

I submit that Dependabot and Renovate Bots likely make 70%+ of the most meaninful supply chain security contributions in both open and closed source.

My preference is to define a contributor as a human or system actor that contributes to the project.

SecurityCRob:
So would a proposed change orbit (get it?  ar ar ar!) around an actor (person vs. robot) or specific actions we desire to better wrangle or perhaps the use af the technology?  I agree with Eddie's early comment that we probably need to get user, maintainer, etc. documented in our docs.

funnelfiasco:
> My preference is to define a contributor as a human or system actor that contributes to the project.

@trumant would you be willing to open a PR so that we have something specific to work against?

trumant:
I consider this issue appropriately addressed for the time being and suggest that @funnelfiasco consider closing it.

[...]

PR #304, opened 2025-05-09
docs: clarify that contribs come from humans and non-humans

trumant:
Adding language to clarify that non-human actors play a role in the SDLC and that those activities are treated equally to human actor activities within the context of the baseline

This updates https://github.com/ossf/security-baseline/issues/275

### DIFF
--- a/baseline/lexicon.yaml
+++ b/baseline/lexicon.yaml
@@ -75,9 +75,9 @@
 - term: Contributor
   definition: |
     Entities who commit code or documentation to
-    the project. Code contributors include
-    collaborators or external participants who
-    submit changes.
+    the project. This includes both human
+    and non-human actors and makes no distinctions
+    based on their role within the project.
 
     In the context of the Open Source Project
     Security Baseline, code contributors does not
@@ -94,10 +94,9 @@
     the primary deliverable in a release.
 - term: Collaborator
   definition: |
-    A user with a role on the project's version
-    control system who can approve changes or
-    manage the repository settings. Collaborators
-    may have varying permission levels based on
+    A human or non-human entity with permissions to
+    approve changes or manage the repository settings.
+    Collaborators may have varying permission levels based on
     their role in the project. This does not
     include contributors whose changes only
     originate through a request from a repository
--- a/baseline/lexicon.yaml
+++ b/baseline/lexicon.yaml
@@ -448,6 +448,14 @@
     recommended formats are Semantic Versioning
     or Calendar Versioning.
   synonyms:
+- term: User
+  definition: |
+    A human making use of project resources, such as
+    the software, documentation, or other community
+    resources. This includes both end-users and
+    contributors.
+  synonyms:
+    - Person
 - term: Version Control System
   definition: |
     A tool that tracks changes to files over time
--- a/baseline/OSPS-AC.yaml
+++ b/baseline/OSPS-AC.yaml
@@ -10,7 +10,7 @@ controls:
   - id: OSPS-AC-01
     title: |
       The project's version control system MUST require multi-factor
-      authentication for collaborators modifying the project repository
+      authentication for users modifying the project repository
       settings or accessing sensitive data.
     objective: |
       Reduce the risk of account compromise or insider threats by requiring


david-a-wheeler (dismissed):

david-a-wheeler:
I agree. Dependabot, etc., are contributors.

eddie-knight:
My original intent here was to have contributors be human or machine, and collaborators as human-exclusive.

I could see an argument that something like Minder is a collaborator, since it modifies sensitive settings... But I'm not entirely convinced.

baseline/lexicon.yaml
```diff
@@ -94,10 +94,9 @@
     the primary deliverable in a release.
 - term: Collaborator
   definition: |
-    A user with a role on the project's version
-    control system who can approve changes or
-    manage the repository settings. Collaborators
-    may have varying permission levels based on
+    A human or non-human entity with permissions to
```
trumant, line 97: Security tooling operating on the repository may have permissions to approve changes and manage repository settings.

I also expect many projects have change approval bots that may be permitted to operate without human change approval intervention.

funnelfiasco: The only issue I see is in [OSPS-AC-01](https://baseline.openssf.org/versions/devel#osps-ac-01---the-projects-version-control-system-must-require-multi-factor-authentication-for-collaborators-modifying-the-project-repository-settings-or-accessing-sensitive-data):

It's not clear how 

> When a user attempts to access a sensitive resource in the project's version control system, the system MUST require the user to complete a multi-factor authentication process.

works with a non-human actor.

trumant: I didn't make updates there because of this language:

> When a user attempts to access a [sensitive resource](https://baseline.openssf.org/versions/devel#sensitive-resource) in the project's [version control system](https://baseline.openssf.org/versions/devel#version-control-system), the system MUST require the user to complete a [multi-factor authentication](https://baseline.openssf.org/versions/devel#multi-factor-authentication) process.

"user" == human in my mind

trumant: We could certainly speak to how non-human actors must authn/authz, but I didn't read the existing requirement as covering that.

funnelfiasco: I also interpret "user" as explicitly "human", but we don't define that. Maybe we just change "collaborators" to "users" in the recommendation for AC-01? That way we're using consistent wording. Or else we can add a definition of user that makes it explicitly clear that it's a human actor.

eddie-knight: The intent of using "collaborator" in AC-01 stems from the intent that collaborator == human contributor with elevated permissions.

If we are changing collaborator to include non-human, then I agree that it should be clear that this is discussing human actors.

On an aside... The more we debate this, the more I question whether a non-human actor performing highly-sensitive actions should be tolerated as safe/secure.

trumant: Does https://github.com/ossf/security-baseline/pull/304/commits/432faa8e28e4eacbe997353b170b1603d7315ec8 help?

funnelfiasco (approved):

david-a-wheeler:
The *final* decisions should be done by humans. So for example, I think "maintainer" should be clearly human, since they make the decision on whether or not a contribution is accepted.

A contributor can be a non-human. E.g., I would say that every PR from dependabot is a contribution from a non-human contributor.

trumant:
> The final decisions should be done by human

For all levels?

I feel like we need to leave room for automated policy, developed and administered by humans, that is the sole approver of a class of changes.

eddie-knight:
How do we have "developed and administered by humans" while also discussing machine-executed changes of sensitive elements? (Or, what am I missing here?)

SecurityCRob (approved):

evankanderson:
> > The final decisions should be done by human
> 
> For all levels?
> 
> I feel like we need to leave room for automated policy, developed and administered by humans, that is the sole approver of a class of changes.

Not to pull out the IBM meme, but until you can hold a computer responsible for contributing code which belongs to someone who does not want it contributed, I think the final decision for code contribution needs to come from a human who is responsible (in the legal sense) for the outcome.

PR #310, opened 2025-05-14
docs: provide more details on the frameworks mapped

trumant:
This change adds descriptions for all of the frameworks and modifies the documentation template to display the framework details in a table.

### Example

Screenshot 2025-05-14 at 6 39 31 PM


eddie-knight (approved):
yes, yes, a hundred times, yes!

Thank you for connecting the dots to the layer2 metadata.

@SecurityCRob or @funnelfiasco could you please proofread the description text? I only gave it a quick skim, seems fine to me.

cmd/go.mod
```diff
@@ -7,7 +7,7 @@ toolchain go1.24.1
 require (
 	github.com/defenseunicorns/go-oscal v0.6.2
 	github.com/goccy/go-yaml v1.17.1
-	github.com/revanite-io/sci v0.3.5-0.20250506145353-e4f84ef5e40b
+	github.com/revanite-io/sci v0.3.6
```
funnelfiasco, line 10: This and the go.sum change are in #307, which is awaiting a second maintainer approval. Practically, it doesn't matter which PR wins the race so I'm not asking you to change this one, but in general let's keep PRs to one logical change.

cmd/template.md
```diff
@@ -127,7 +127,10 @@ For more information on the project and to make contributions, visit the [GitHub
 
 Controls within this document may map to the following external frameworks:
 
-{{ range .Frameworks }}- [{{ .Title }} ({{ .ID }}): {{ .Version }}]({{ .URL }})
+| ID | Title | Version | Description |
```
funnelfiasco, line 130: I'm not sure this is going to render very well, for the same reasons expressed in #99. Can you publish this on your fork (`git push origin HEAD:main --force` will do the trick, assuming you have Pages configured on your fork) so that we can see what it looks like in practice? We might want to consider a different layout.

funnelfiasco (commented):
The descriptions look fine to me. My main concern is the tabular presentation. I won't nack it because if it gets merged and we don't like the tables, we can simply fix them, but I'd like to know how they'll work before I explicitly ✅ it.

funnelfiasco:
> The descriptions look fine to me. My main concern is the tabular presentation. I won't nack it because if it gets merged and we don't like the tables, we can simply fix them, but I'd like to know how they'll work before I explicitly ✅ it.

I guess it would help if I looked at comment 0. @trumant is that a screen shot of Markdown rendered by GitHub or is it the Jekyll-rendered markdown we'd see on the web page?

cmd/template.md
trumant: @funnelfiasco Great point and thank you for the guidance. See https://trumant.github.io/security-baseline/versions/devel#external-frameworks

trumant:
> > The descriptions look fine to me. My main concern is the tabular presentation. I won't nack it because if it gets merged and we don't like the tables, we can simply fix them, but I'd like to know how they'll work before I explicitly ✅ it.
> 
> I guess it would help if I looked at comment 0. @trumant is that a screen shot of Markdown rendered by GitHub or is it the Jekyll-rendered markdown we'd see on the web page?

The screenshot above was from my VS Code Markdown Preview. I should have specified that.

cmd/template.md
funnelfiasco: It seems fine enough on a desktop, but looking at it from a phone/tablet window, the table is either smushed or requires a lot of sideways scrolling.

Two approaches we could take:

## Approach 1

````
- {{ .ID }}: [{{ .Title}}]({{ .Url }}) version {{  .Version }}
    - {{ .Description }}
````

## Approach 2 

````
| ID | Title | Version |
|----|-------|---------|-------------|
| {{ .Id }} | [{{ .Title }}]({{ .Url }}) | {{ .Version }}  |
````

Approach 1 would still be funky for mappings with a long title (800-161, I'm looking at you), but we could reasonably trim those if needed. I don't looove the idea of an unordered list here, so we could use some `<div>`s instead to make it more a set of indented-paragraphs. That adds complexity, though.

Approach 2 would make the table cleaner, at the cost of explaining what the dang things are. I suppose there's an argument to be made that if anyone cares about the mapping, they already know what it is.

## Approach 3 

A third approach would be to not dynamically produce the full listing. Instead we can make it a static page in `docs/` as supplemental documentation, and the listing on the actual Baseline version page could be something like

````
- [{{ .ID }}](link to docs/mappings.md with the appropriate anchor) version {{ .Version }}
````

With the 3rd approach, we have a static "OSPS Baseline version X maps to version Y of other standard Z", but we can present the expanded explanation on a static page that we can be a little more flexible about formatting.

My inclination, I think, is to go with approach 3, but I'm not strongly tied to it. What are your thoughts?

trumant: > A third approach would be to not dynamically produce the full listing. Instead we can make it a static page in docs/ as supplemental documentation

Existing content in one giant page vs separate page and static vs generated are two different concerns.

If we want to split out content into multiple pages, I'm a big supporter. I would imagine this leading to:

- `docs/versions/${VersionID}/index.md` - contains overview and existing controls
- `docs/versions/${VersionID}/lexicon.md` - contains all of our terms and defs
- `docs/versions/${VersionID}/compliance_mappings.md` - contains all of the frameworks with detailed descriptions. _Possibly also includes a reverse mapping under each framework, showing the Baseline control IDs for controls that map to the framework.

Finally, if we do split out content, my strong preference is to continue to generate said content from the data in the baseline YAML to the maximal extent practical.

funnelfiasco: I'd push back on what you're suggesting and make neither the lexicon nor the mappings (other than a version of the Baseline is tied to a version of an external standard) tied to a particular version. Those are both more along the lines of supplemental and static (in terms of not changing) documentation that supports the Baseline. The fact that we put them all into one file based on a version is an result of how we designed the tooling, not of the intended outcome.

funnelfiasco: At any rate, we're getting pretty far afield from your original proposal. So let's ignore approach 3 for the moment (I think it's the best end result, but clearly there are some discussions to have about how we get there). Given your intent with adding the descriptions (which I think makes sense), it seems like approach 1 is the best short-term answer?

I'm also willing to approve this as-is and then we can discuss a reformat later if you don't want to make any changes right now. Nothing in this commits us to an approach in perpetuity.

trumant: > willing to approve this as-is and then we can discuss a reformat later if you don't want to make any changes right now. Nothing in this commits us to an approach in perpetuity

Sounds good to me

funnelfiasco (approved):
Approving, although I think we'll want to change the presentation approach down the road.

I'll wait until tomorrow to merge to give others time to weigh in as well.

PR #311, opened 2025-05-15
fix: render the framework mapping references in docs

trumant:
Visiting https://baseline.openssf.org/versions/devel#osps-ac-0101 shows that the content under the heading `External Framework Mappings` is missing from the control content in the page.

This change adjusts the field naming of the baseline controls in the data files to match the expected name of `guideline-mappings` and adjusts the template file slightly to match the expected sci/layer2 types and fields available for templating.

This is a defect that should have been caught by our CI process, but was not due to the lack of strict unmarshaling configuration in `loadControlFamily`. This change adds that configuration which results in the following failure against the `main` branch when running `cd cmd && go run . compile --output foo.md`:

```
ERROR: loading control family AC: error decoding AC YAML: [19:5] unknown field "mappings"
  16 |       Reduce the risk of account compromise or insider threats by requiring
  17 |       multi-factor authentication for collaborators modifying the project
  18 |       repository settings or accessing sensitive data.
> 19 |     mappings:
           ^
  20 |       - reference-id: BPB
  21 |         identifiers:
  22 |           - CC-G-1
  23 |       
exit status 1
```

### DIFF
--- a/baseline/OSPS-AC.yaml
+++ b/baseline/OSPS-AC.yaml
@@ -16,7 +16,7 @@ controls:
       Reduce the risk of account compromise or insider threats by requiring
       multi-factor authentication for collaborators modifying the project
       repository settings or accessing sensitive data.
-    mappings:
+    guideline-mappings:
       - reference-id: BPB
         identifiers:
           - CC-G-1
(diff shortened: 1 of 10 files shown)


funnelfiasco (approved):
This is a result of #284, I assume. Looks good to me.

eddie-knight (approved):

PR #333, opened 2025-06-02
Update faq.md - What is in scope for the Open Source Project Security Baseline, and what is out of scope ? 

victorjunlu:
Worked together with @eddie-knight on this change.

What is in scope for the Open Source Project Security Baseline, and what is out of scope ?

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -73,6 +73,12 @@ Since the Baseline is designed for the developers of a project, not the consumer
 OSPS Baseline compliance is a point-in-time status.
 We encourage projects using the OSPS Baseline to say something like “As of April 31, 2025, this project complies with OSPS Baseline version 2025-02-30 level 2.”
 
+## What is in scope for the Open Source Project Security Baseline, and what is out of scope?
+
+This baseline seeks to address security hygiene elements — those which lock down the ways of working, delivering the product, and equipping its users to adopt it safely.  To use an analogy, it’s like home builders who lock up tools, secure the construction site, and control who enters, but also ensure the finished house is handed over with clear instructions for safe use. It’s not about changing the blueprint—it’s about protecting the build process and delivering a home that’s ready to live in securely.
+
+By contrast, secure design and development are out of scope for this activity. Continuing the analogy, those activities would be the responsibility of the architects and builders who create the blueprint and decide how the house is constructed to prevent break-ins—with reinforced doors, secure locks, strategic layouts/no backdoors, or built-in security systems. It’s about designing security into the structure itself, not just safeguarding the build and handoff. 
+
 ## How can I get involved in the OSPS Baseline project?
 The OSPS Baseline project welcomes contributions in the [GitHub repository](https://github.com/ossf/security-baseline/pull/24/files).
 For discussion, join us in [#sig-security-baseline](https://openssf.slack.com/archives/C07DC6TT2QY) in the OpenSSF Slack instance.


eddie-knight (approved):

funnelfiasco (changes requested):
Thanks for your contribution, @victorjunlu!

Before I get to the substance, a style note: although we don't have a formal style guide yet, you'll note that this file uses a one-line-per-sentence style. As you address other feedback, please update your PR to reflect that.

To the substance: It's not clear to me that "secure development" is out of scope. Or at least, what it means differently from "security hygiene." Looking through the controls, I see several that would seem to fit in your analogy as "out of scope"

* [OSPS-VM-06.02](https://baseline.openssf.org/versions/devel#osps-vm-0602)
* [OSPS-SA-03.02](https://baseline.openssf.org/versions/devel#osps-sa-0302)
* [OSPS-QA-06.03](https://baseline.openssf.org/versions/devel#osps-qa-0603)
* [OSPS-SA-03.01](https://baseline.openssf.org/versions/devel#osps-sa-0301)
* [OSPS-SA-01.01](https://baseline.openssf.org/versions/devel#osps-sa-0101)

I'd also argue that things like dependency selection (which is covered in the OSPS Baseline) falls into the "decide how the house is constructed" part of your analogy.

To me, the analogy _reduces_ clarity.

It could be that the solution is to remove some of the controls because they're out of scope. I'm not sure I agree with that, but I'm open to the argument. But in any case, I don't think this FAQ entry matches the current reality, so one or both of them will need to be adjusted.

eddie-knight:
I'm not attached to this FAQ entry, but I do think it's accurate.

- OSPS-VM-06.02: Is this policy in place?
- OSPS-SA-01.01: Have you documented your design?
- OSPS-SA-03.01: Have you done a security assessment?
- OSPS-SA-03.02: Have you done a threat model?
- OSPS-QA-06.03: Is this policy in place?

These items don't evaluate efficacy or content, just that a best practice has been done. 

ie, We're not checking the quality of your blueprint... just making sure that you have one on record with the city.

docs/faq.md
```diff
@@ -73,6 +73,12 @@ Since the Baseline is designed for the developers of a project, not the consumer
...
 
+## What is in scope for the Open Source Project Security Baseline, and what is out of scope?
+
+This baseline seeks to address security hygiene elements — those which lock down the ways of working, delivering the product, and equipping its users to adopt it safely.  To use an analogy, it’s like home builders who lock up tools, secure the construction site, and control who enters, but also ensure the finished house is handed over with clear instructions for safe use. It’s not about changing the blueprint—it’s about protecting the build process and delivering a home that’s ready to live in securely.
+
+By contrast, secure design and development are out of scope for this activity. Continuing the analogy, those activities would be the responsibility of the architects and builders who create the blueprint and decide how the house is constructed to prevent break-ins—with reinforced doors, secure locks, strategic layouts/no backdoor, or built-in security systems. It’s about designing security into the structure itself, not just safeguarding the build and handoff.
```
eddie-knight, line 80:
```suggestion
This baseline seeks to address security hygiene elements — those which lock down the ways of working, delivering the product, and equipping its users to adopt it safely.
To use an analogy, it’s like home builders who lock up tools, secure the construction site, and control who enters, but also ensure the finished house is handed over with clear instructions for safe use. 
We're not checking the quality of your blueprint... just making sure that you have one on record with the city.
Design and development are part of the [assessment](https://github.com/ossf/security-assessments) activity, which is comparable to evaluating the blueprint of a home before it is constructed.
```

evankanderson: I'm not sure I like the term "lock down" here (even if we're talking about controls, I prefer to think of it as "defending against yourself on your worst days").  How about:

> The baseline seeks to address security hygiene practices — much like workspace and product safety requirements in other domains.  Using a construction analogy, baseline is similar to the jobsite rules and processes (hard hats, fencing, etc) a builder uses to prevent injuries to workers as they are building a house, and also covers the documentation standards used to provide operation and maintenance instructions for the final owner.

funnelfiasco: I think some combination of Eddie's and Evan's suggestions are the right answer here. Let's see if I can come up with a good blend. :-)

funnelfiasco:
```suggestion
OSPS Baseline seeks to address security hygiene elements — those which secure the ways of working, delivering the product, and equipping its users to adopt it safely.
To use an analogy, Baseline is similar to a foot processing plant requiring practices to ensure safe food products: wearing hair nets, regularly cleaning and sanitizing equipment, monitoring refrigerator temperatures, etc.

OSPS Baseline is not a food critic and does not evaluate the taste of the finished food; that is the role of [OSPS Assessments](https://github.com/ossf/security-assessments).
OSPS Baseline is more like the health department guidelines that requires safe production of food.
```

I think the building analogy is fundamentally flawed, so I propose a food analogy instead. What do we think of this?

PR #334, opened 2025-06-05
chore: update revanite-io/sci from 0.3.6 to 0.3.7

trumant:
This will successfully produce the same result as https://github.com/ossf/security-baseline/pull/332 is trying to


funnelfiasco (changes requested):
@trumant this has additional, unrelated commits. Can you please remove those? It looks like at least some of them are in main, it shouldn't be _too_ much of a problem, but better to be clean about it. Otherwise, this looks good (and the tests pass, so yay!)

PR #342, opened 2025-06-12
Converted lexicon references to clickable links

### LINKED ISSUE #322: Render lexicon references as hyperlinks
It looks all of the references are URLs, so we might as well make them clickable to improve the user friendliness.

joveuh:
Fixes #322


funnelfiasco:
Hi, @joveuh! I appreciate your effort here, but I don't think it's the right approach. First of all, it's not building correctly, as you'll notice:

~~~~
ERROR: error reading lexicon: error decoding YAML: [27:96] value is not allowed in this context
  26 |   references:
> 27 |     - [https://cheatsheetseries.owasp.org/cheatsheets/Attack_Surface_Analysis_Cheat_Sheet.html](https://cheatsheetseries.owasp.org/cheatsheets/Attack_Surface_Analysis_Cheat_Sheet.html)
~~~~

It's possible that you can work around this with some quoting, but that means any new addition will require correctly formatting and quoting a new entry.

I think a better approach would be for the compile command to turn those references into links at build time. There's an [existing template function `asLink`](https://github.com/ossf/security-baseline/blob/main/cmd/pkg/baseline/template-functions.go#L82-L86) that turns things into in-page links. That could be modified to support external links or perhaps a similar function could be added to generate external links.

Also: I noticed that the commit is associated with a different GitHub account than the one that opened the PR. Both accounts use your name, so I'm not overly concerned, but it would be better to have them match. (Perhaps by associating your commit with an email address unique to this account).

funnelfiasco:
@joveuh are you still interested in working on this? If not, I'll close this PR and we can take another pass later.

joveuh:
Hey @funnelfiasco , I haven't gotten to it yet still, let's close it for now and I can loop back later. Thanks ,

funnelfiasco:
Sounds good!

PR #350, opened 2025-07-08
Renamed some framework IDs (non-functional)

eddie-knight:
This is related to, but not dependent on or blocking #349

I noticed that some of the framework names could have clearer IDs, and since we aren't restricted by length, I've proposed:

- `ScCrd` -> `Scorecard`
- `OCRE` -> `OpenCRE`
- `OC` -> `ISO-18974`

For review simplicity, I'd like to do this prior to migrating to the full use of metadata.yaml in a subsequent PR

### DIFF
--- a/.project-words.txt
+++ b/.project-words.txt
@@ -15,7 +15,7 @@ lifecycles
 NCSC
 openssf
 organisations
-OCRE
+OpenCRE
 OSCAL
 OSPS
 PCIDSS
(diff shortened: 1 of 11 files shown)


funnelfiasco (approved):
Looks like a good change to me. We could call `18974` `iso18974` if that feels a little clearer, but I'm fine either way.

I built it in my fork and everything worked as expected.

SecurityCRob (dismissed):

funnelfiasco:
Whoops! Fixing a conflict invalidated previous reviews.

trumant (approved):

SecurityCRob (approved):

PR #356, opened 2025-07-11
Add DO-01.02 disclaimer requirement

### LINKED ISSUE #353: Create a control for a statement on supported use cases
In discussing #302 today:

* This feels like an extension of DO-01
* Need to think about where our boundary is in terms of “how to use the software securely” in terms of maintainer
* Decision: @evankanderson proposes a statement on the software indicating its intended use cases

evankanderson:
Fixes #353

Given that this is attempting to disprove a negative (the software is NOT intended for general use), I'm not sure how we can automatically validate this.

An example of software that might be flagged by this would be https://github.com/sloria/doitlive -- given that this is running an arbitrary shell script, there should probably be a disclaimer to \_trust\_ the shell script you're running it with (reasonable, but might not be obvious given the propensity of `curl | bash -` installs...).


baseline/OSPS-DO.yaml
```diff
@@ -75,6 +75,20 @@ controls:
           project, explaining how to install, configure, and use the project's
           features. If there are any known dangerous or destructive actions
           available, include highly-visible warnings.
+      - id: OSPS-DO-01.02
+        text: |
```
trumant, line 85: The underlying intent here seems to have some parallels with CISA's Secure by Design guidance in the "Reduce hardening guide size" section of https://www.cisa.gov/sites/default/files/2023-10/SecureByDesign_1025_508c.pdf

Specifically, I think we could improve this by changing the text to something like:

```suggestion
        text: |
          The project documentation MUST include guidance on secure usage of the project.
        applicability:
          - Maturity Level 2
          - Maturity Level 3
        recommendation: |
          Document the security-relevant usage and configuration options of the project clearly in a hardening or security guide that is prominently noted in the installation or usage documentation.
```

Examples of such guidance documents:

- https://v2.helm.sh/docs/securing_installation/
- https://theupdateframework.io/docs/security/
- https://www.openpolicyagent.org/docs/security

WDYT?

trumant: From the Secure by Design guidance:

> Track and reduce “hardening guide” size. Reduce the size of “hardening guides”
that are included with products and strive to ensure that the size shrinks over time as
new versions of the sofware are released Integrate components of the “hardening
guide” as the default configuration of the product The authoring organizations
CISA | NSA | FBI | ACSC | CCCS | CERT NZ | NCSC-NZ | NCSC-UK | BSI | NCSC-NL
NCSC-NO | NÚKIB | INCD | KISA | NISC-JP | JPCERT/CC | CSA | CSIRTAMERICAS recognize that shortened hardening guides result from ongoing partnership with existing customers and include eforts by many product teams, including user experience (UX)

> • Consider the user experience consequences of security setings. Each new seting
increases the cognitive burden on end users and should be assessed in conjunction
with the business benefit it derives Ideally, a seting should not exist; instead, the most
secure seting should be integrated into the product by default When configuration is
necessary, the default option should be broadly secure against common threats

eddie-knight: I think these are good thoughts, but potentially additive.

I think the intent is to signal to users that the project outputs are not ready for sensitive environments.

My thought was that we could have a boolean field in security insights, and potentially extend that check with a scan for boilerplate text in the README.

Although if my understanding is correct, @evankanderson, I'd be curious to see a recent example of this in practice.

.gitignore
funnelfiasco, line 1: This and `CONTRIBUTING.md` changes should be in a separate PR for clarity (although they are both good changes!)

funnelfiasco:
@evankanderson are you still interested in pursuing this?

SecurityCRob (approved):

PR #358, opened 2025-07-18
Disambiguate version control system and repository

### LINKED ISSUE #66: Disambiguating version control system and repository
In OSPS-01:

> The project's [version control system](https://github.com/ossf/security-baseline/blob/main/baseline.md#version-control-system) MUST require [multi-factor authentication](https://github.com/ossf/security-baseline/blob/main/baseline.md#multi-factor-authentication) for [collaborators](https://github.com/ossf/security-baseline/blob/main/baseline.md#collaborator) modifying the project [repository](https://github.com/ossf/security-baseline/blob/main/baseline.md#repository) settings or accessing sensitive data.

The definition for VCS points to examples like Git that cannot enforce multi-factor authentication. The definition for "repository" is closer to a source control platform (SCP) such as GitHub or GitLab that can enforce multi-factor authentication. I'm unsure if just replacing VCS with repository is sufficient, especially since the word "repository" doesn't always mean the centrally hosted copy to everyone.

Separately, does this requirement exclude other options for hosting source code? For example, I'm unsure if https://git.kernel.org/ can meet the MFA requirement.

funnelfiasco:
Fixes #66

### DIFF
--- a/baseline/OSPS-AC.yaml
+++ b/baseline/OSPS-AC.yaml
@@ -72,8 +72,8 @@ controls:
     assessment-requirements:
       - id: OSPS-AC-01.01
         text: |
-          When a user attempts to access a sensitive resource in the project's
-          version control system, the system MUST require the user to complete
+          When a user attempts to read or modify a sensitive resource in the project's
+          central repository, the system MUST require the user to complete
           a multi-factor authentication process.
         applicability:
           - Maturity Level 1


eddie-knight (approved):

baseline/OSPS-AC.yaml
```diff
@@ -72,8 +72,8 @@ controls:
     assessment-requirements:
       - id: OSPS-AC-01.01
         text: |
-          When a user attempts to access a sensitive resource in the project's
-          version control system, the system MUST require the user to complete
+          When a user attempts to read or modify a sensitive resource in the project's
+          central repository, the system MUST require the user to complete
```
trumant: By "central" are you implying "authoritative"? I.e. an "official project channel" as mentioned in https://baseline.openssf.org/versions/devel#osps-br-0301

As written, I think its clear, but if we can be even more explicit and/or use language we use in other requirements and further improve clarity, I'm for it.

funnelfiasco: Hm. Yeah, I think "authoritative" works here. I was thinking more "centralized", but "authoritative" gets to the deeper meaning, IMO.

puerco (approved):

baseline/OSPS-AC.yaml
eddie-knight: I kinda squinted at this part... I think it's clear enough, but nitpicking could raise questions about multi-repo projects as well as the definition of central/authoritative. A lexicon entry might help for the terminology.

SecurityCRob (approved):
i'm good with the suggested change

PR #367, opened 2025-08-05
Convert docs theme to minima and add template elements for baseline site

### LINKED ISSUE #341: Update GitHub pages template
As mentioned in [today's meeting](https://docs.google.com/document/d/16tL1Ln7owIRXSoCKgyYHCs9-JP9iw-ouyk8koGAeHA0/edit?tab=t.0#heading=h.szjt93srozbw), Jeff asked if we would like GitHub Pages template updates applied to the Baseline content (header & footer with logo, etc). We would.

GeauxJD:
Resolves #341 

Updates to Jekyll's default minima theme
Adds header and footer content to templates
Updates to OpenSSF brand guidelines fonts
adds front-matter to 3 of the markdown pages for navigation functionality
Corrects a minor grammar issue in index.md


.gitignore
```diff
@@ -3,3 +3,6 @@ docs/versions/devel.md
 
 # generated output from go run ./... compile
 checklist.md
+Gemfile
+Gemfile.lock
```
funnelfiasco: Do we want to ignore these? Seems like it'd be beneficial for publishing (and for folks doing local builds, too)

docs/_config.yml
```diff
@@ -1,12 +1,79 @@
...
-title: null
+title: Open Source Project Security Baseline
+author:
+  name: OpenSSF Security Baseline SIG
+  email: openssf-sig-security-baseline+subscribe@lists.openssf.org
+logo: /assets/logo.svg
+url: "https://baseline.openssf.org"
+description: > 
+  The Open Source Project Security (OSPS) Baseline is a set of security 
+  criteria that projects should meet to demonstrate a strong security posture.
```
funnelfiasco:
```suggestion
  controls that projects should meet to demonstrate a strong security posture. 
```

This makes the wording consistent (although we can also fix it later)

funnelfiasco (changes requested):
This looks pretty good overall. I think we'll want to include the Gemfile* ...files, though. We should also update the spellcheck config so it doesn't complain about the non-words that it's complaining about (I wish it were better about distinguishing things here, but alas)

Once we merge this, I'll need to update the release instructions to include modifying the current version to have the `nav-title`.

evankanderson:
> This looks pretty good overall. I think we'll want to include the Gemfile* ...files, though. We should also update the spellcheck config so it doesn't complain about the non-words that it's complaining about (I wish it were better about distinguishing things here, but alas)

We could spell-check only `.md` and `.yaml` files: https://github.com/streetsidesoftware/cspell-action?tab=readme-ov-file#usage

> 
> Once we merge this, I'll need to update the release instructions to include modifying the current version to have the `nav-title`.

docs/versions/2025-02-25.md
```diff
@@ -1,3 +1,7 @@
+---
+nav-title: Current Version
```
evankanderson, line 2:
```suggestion
nav-title: 2025-02-25 (Current)
```

.gitignore
evankanderson: https://baseline.openssf.org/versions/2025-02-25#osps-qa-0201 suggests that we should probably have at least the Gemfile (and I'd recommend both)

eddie-knight: That's an interesting take... I don't read it that way. Since QA-02 is about "the software's direct dependencies," I wouldn't expect that dependencies for the documentation website would be covered by the test requirement.

docs/versions/2025-02-25.md
GeauxJD: Not sure if the 2025-02-25 as the title in a navbar link is going to be very intuitive for someone who is new to the project

GeauxJD:
> > This looks pretty good overall. I think we'll want to include the Gemfile* ...files, though. We should also update the spellcheck config so it doesn't complain about the non-words that it's complaining about (I wish it were better about distinguishing things here, but alas)
> 
> We could spell-check only `.md` and `.yaml` files: https://github.com/streetsidesoftware/cspell-action?tab=readme-ov-file#usage

Can I get some help from someone on the spellcheck piece? I'm not familiar with cspell or how to configure it.

docs/versions/2025-02-25.md
funnelfiasco: I agree with Jeff. We could do "Current Version (2025-02-05)" instead, but I think for navbar purposes "Current version" is sufficient.

.gitignore
funnelfiasco: I agree that the docs website isn't strictly covered by QA-02, but the spirit of QA-02 applies. And, if nothing else, it makes life a lot more convenient for local builds, etc.

funnelfiasco (dismissed):
Looks good to me. Thanks @GeauxJD!

docs/Gemfile
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+source "https://rubygems.org"
+
+# gem "jekyll"
+gem "github-pages", "~> 232", group: :jekyll_plugins
+# gem "rails"
```
trumant, line 7:
```suggestion
gem "github-pages", "~> 232", group: :jekyll_plugins
```

trumant (approved):

.gitignore
evankanderson: If the output (released artifact) is the website, then the Jekyll configuration (and the go templating) are both the software used to produce the output.  If that's not the intent, then we should figure out how to clarify.

On the other hand, the version of Jekyll used to generate the website is actually specified by https://github.com/ossf/security-baseline/blob/main/.github/workflows/web-publish.yml#L36, so maybe this is just "development cruft" that could be ignored as it only impacts contributors attempting to build a local copy.

docs/_layouts/default.html
```diff
@@ -1,32 +1,27 @@
...
+<html lang="{{ site.lang | default: "en-US" }}">
   <head>
     <meta charset="UTF-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width, initial-scale=1">
 
 {% seo %}
-    <link rel="stylesheet" href="{{ "/assets/css/style.css?v=" | append: site.github.build_revision | relative_url }}">
-    {% include head-custom.html %}
+        <link rel="stylesheet" href="{{ '/assets/css/style.css' | relative_url }}">
```
evankanderson: I don't see where `custom-head.html` is being included now?

Also, it looks like the previous stylesheet link had some cache-busting parameters for when the CSS was updated.  I don't care strongly, but further updates to the CSS may take longer to show up (and I'm fine with that).

docs/_layouts/default.html
```diff
@@ -1,32 +1,27 @@
...
+    <!--[if lt IE 9]>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/html5shiv/3.7.3/html5shiv.min.js"></script>
+    <![endif]-->
```
evankanderson: If you're running IE < 9.0, you're running a browser from 2009 or earlier, and it's even [out of support on Windows Embedded POSReady 2009 with extended support updates](https://en.wikipedia.org/wiki/Internet_Explorer_version_history#OS_compatibility).  I think you can drop this.

docs/assets/fonts/NOTES.txt
```diff
@@ -0,0 +1,14 @@
...
+    https://openssf.org/brand-guidelines/ 
+
+* https://fonts.google.com/specimen/IBM+Plex+Sans
+  - Licensed under the Open Font License (OFL)
+  - IBM Plex Sans is used for headlines per
+    https://openssf.org/brand-guidelines/ 
+
+* https://fonts.google.com/specimen/Teko
+  - This *was* the recommended heading font, but is no longer.
+    Many found it hard to read.
```
evankanderson: We don't have this font -- remove this?

evankanderson (approved):
A few minor nits, but I'm willing to merge as-is.

docs/_layouts/default.html
GeauxJD: i think you are right that custom-head.html is a legacy file from copying some things in from a setup of these templates elsewhere. we can probably do a follow-on issue to test removing that file and confirm nothing breaks.

docs/_layouts/default.html
GeauxJD: Yeah this came straight from the minima theme defaults, i'd be ok with us testing removal in a follow-on issue

docs/assets/fonts/NOTES.txt
GeauxJD: Good catch, can remove this. @funnelfiasco let me know if you want me to do some more commits now to address some of these nits or if we want to followup with another PR later for cleanups.

funnelfiasco: Might as well address them now, unless you want to juice your stats :-D

SecurityCRob (approved):
i approve of these updates

funnelfiasco:
@GeauxJD did you want to address the other comments in this PR or should I merge it and create a new issue? Either way is fine with me, but if it'll be a bit before you can get to the outstanding parts, I'd like to land the work you've done so far.

docs/_layouts/default.html
GeauxJD: Removed conditional IE9 script

docs/_layouts/default.html
GeauxJD: removed custom-head file that is no longer used

docs/assets/fonts/NOTES.txt
GeauxJD: Removed reference to the font in notes

PR #373, opened 2025-08-13
feat: add control OSPS-BR-07 for secrets management

### LINKED ISSUE #352: Add a control for secrets management
In discussing #302 today,

* Need a “don’t have secrets in your repo”
* “Don’t put secrets in a Google Doc”
* Decision: Need some guidance around using appropriate secrets storage solution
* Sample wording from Eddie: “When a development tool uses secrets for authentication … uses a secure “

trumant:
This change adds a new control to the Build & Release control family that focuses on secure handling and storage of project secrets.

This change closes #352

### DIFF
--- a/baseline/OSPS-BR.yaml
+++ b/baseline/OSPS-BR.yaml
@@ -490,7 +490,7 @@ controls:
         applicability:
           - Maturity Level 3
         recommendation: |
-          Document how secrets and credentials are managed and used within the project. This should include details on how secrets are stored (e.g., using a secrets management tool), how access is controlled, and how secrets are rotated or updated. Ensure that sensitive information is not hard-coded in the source code or stored in version control systems.
+          Document how secrets and credentials are managed and used within the project. This should include details on how secrets are stored (e.g., using a secrets management tool), how access is controlled, and how secrets are rotated or updated.
       - id: OSPS-BR-07.02
         text: |
           The project MUST prevent the unintentional storage of unencrypted sensitive data, such as secrets and credentials, in the version control system.
--- a/baseline/OSPS-BR.yaml
+++ b/baseline/OSPS-BR.yaml
@@ -469,7 +469,7 @@ controls:
     title: |
       The project MUST store and manage all secrets and credentials used by the project in a secure manner.
     objective: |
-      Ensure that sensitive data is not disclosed, compromised or misused leading to security vulnerabilities or data breaches.
+      Ensure that sensitive data is not disclosed, compromised or misused leading to security vulnerabilities or supply chain compromise.
     guideline-mappings:
       - reference-id: BPB
         identifiers:
--- a/baseline/OSPS-BR.yaml
+++ b/baseline/OSPS-BR.yaml
@@ -486,15 +486,15 @@ controls:
     assessment-requirements:
       - id: OSPS-BR-07.01
         text: |
-          The project MUST define a policy for managing secrets and credentials used by the project. The policy should include guidelines for storing, accessing, and rotating secrets and credentials.
-        applicability:
-          - Maturity Level 3
-        recommendation: |
-          Document how secrets and credentials are managed and used within the project. This should include details on how secrets are stored (e.g., using a secrets management tool), how access is controlled, and how secrets are rotated or updated.
-      - id: OSPS-BR-07.02
-        text: |
           The project MUST prevent the unintentional storage of unencrypted sensitive data, such as secrets and credentials, in the version control system.
         applicability:
           - Maturity Level 1
         recommendation: |
-          Configure .gitignore or equivalent to exclude files that may contain sensitive information. Use pre-commit hooks and automated scanning tools to detect and prevent the inclusion of sensitive data in commits. Periodically audit the version control history to ensure that no sensitive information has been inadvertently committed.
+          Configure .gitignore or equivalent to exclude files that may contain sensitive information. Use pre-commit hooks and automated scanning tools to detect and prevent the inclusion of sensitive data in commits.
+      - id: OSPS-BR-07.02
+        text: |
+          The project MUST define a policy for managing secrets and credentials used by the project. The policy should include guidelines for storing, accessing, and rotating secrets and credentials.
+        applicability:
+          - Maturity Level 3
+        recommendation: |
+          Document how secrets and credentials are managed and used within the project. This should include details on how secrets are stored (e.g., using a secrets management tool), how access is controlled, and how secrets are rotated or updated. Ensure that sensitive information is not hard-coded in the source code or stored in version control systems.
--- a/baseline/OSPS-BR.yaml
+++ b/baseline/OSPS-BR.yaml
@@ -465,3 +465,36 @@ controls:
           signature or attestations, such as GPG or PGP signature, Sigstore
           signatures, SLSA provenance, or SLSA VSAs. Include the cryptographic
           hashes of each asset in a signed manifest or metadata file.
+  - id: OSPS-BR-07
+    title: |
+      The project MUST store and manage all secrets and credentials used by the project in a secure manner.
+    objective: |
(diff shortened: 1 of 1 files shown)


baseline/OSPS-QA.yaml
```diff
@@ -545,3 +545,36 @@ controls:
...
           be merged.
+  - id: OSPS-QA-08
+    title: |
+      The project MUST store and manage all secrets and credentials used by the project in a secure manner.
+    objective: |
+      Ensure that sensitive data is not disclosed, compromised or misused leading to security vulnerabilities or data breaches.
+    guideline-mappings:
+      - reference-id: BPB
+        identifiers:
+          - S-B-5 # TODO: is this the right numbering for https://www.bestpractices.dev/en/criteria#0.no_leaked_credentials
```
eddie-knight: Also, we have an outstanding action item to go back through all of the BPB identifiers to make them match the origin ID instead of the custom format

baseline/OSPS-QA.yaml
```diff
@@ -545,3 +545,36 @@ controls:
...
+          - P0.3.1
+          - P0.4.2
+          - PO.5.1
+          - PW.1.2
+          - PW.1.3
+          - PW.5.1
+    assessment-requirements:
+      - id: OSPS-QA-08.01
+        text: |
+          The project MUST define a policy for managing secrets and credentials used by the project. The policy should include guidelines for storing, accessing, and rotating secrets and credentials.
```
eddie-knight: Do we need a caveat for any situations where the project does not manage its own secrets? 

For example, some FINOS projects don't have anything that requires secrets. And in cases where cloud resources are used in the test pipeline, FINOS manages cloud credentials outside of the project's scope.

trumant: I believe the language above is inclusive enough to account such arrangements. I.e. in these cases, the project could say "Our policy is to never manage any secrets/credentials related to the project's application source code and CI/CD usage, instead those are managed by (those other folks over there)"

funnelfiasco:
```suggestion
          The project MUST define a policy for managing secrets and credentials used by the project which includes guidelines for storing, accessing, and rotating secrets and credentials.
```

Avoid "should"

funnelfiasco:
Thinking about this a little more, I don't think "Quality" is the right section. That's described as:

> Quality focuses on the processes and practices used to ensure the quality and reliability of the project's source code and software assets. These controls help ensure that the project's source code is well maintained, secure, and reliable, reducing the risk of defects or vulnerabilities in the software.

There's no perfect fit for this one, but I'd say "Access Control" is probably the closest?

> Access Control focuses on the mechanisms and policies that control access to the project's version control system and CI/CD pipelines. These controls help ensure that only authorized users can access sensitive data, modify repository settings, or execute build and release processes.

evankanderson (commented):
Should this be in Build & Release, since that is where most of the secrets are used?

funnelfiasco (dismissed):

funnelfiasco (dismissed):

baseline/OSPS-BR.yaml
```diff
@@ -465,3 +465,36 @@ controls:
           signature or attestations, such as GPG or PGP signature, Sigstore
           signatures, SLSA provenance, or SLSA VSAs. Include the cryptographic
           hashes of each asset in a signed manifest or metadata file.
+  - id: OSPS-BR-07
+    title: |
+      The project MUST store and manage all secrets and credentials used by the project in a secure manner.
+    objective: |
+      Ensure that sensitive data is not disclosed, compromised or misused leading to security vulnerabilities or data breaches.
```
evankanderson: I'm not sure if "data breaches" is the right term here.  Maybe "supply chain compromise"?


```suggestion
      Ensure that sensitive data is not disclosed, compromised or misused leading to security vulnerabilities or supply chain compromise.
```

baseline/OSPS-BR.yaml
```diff
@@ -465,3 +465,36 @@ controls:
...
+          - PW.1.3
+          - PW.5.1
+    assessment-requirements:
+      - id: OSPS-BR-07.01
+        text: |
+          The project MUST define a policy for managing secrets and credentials used by the project. The policy should include guidelines for storing, accessing, and rotating secrets and credentials.
+        applicability:
+          - Maturity Level 3
+        recommendation: |
+          Document how secrets and credentials are managed and used within the project. This should include details on how secrets are stored (e.g., using a secrets management tool), how access is controlled, and how secrets are rotated or updated. Ensure that sensitive information is not hard-coded in the source code or stored in version control systems.
```
evankanderson:
```suggestion
          Document how secrets and credentials are managed and used within the project. This should include details on how secrets are stored (e.g., using a secrets management tool), how access is controlled, and how secrets are rotated or updated.
```

The last sentence seems duplicative of the .02 control.

baseline/OSPS-BR.yaml
```diff
@@ -465,3 +465,36 @@ controls:
...
+      - id: OSPS-BR-07.02
```
evankanderson, line 494: Do we want to reverse the numbering of these controls so that .01 is in level 1 and .02 is in level 3?

evankanderson (approved):
The recommendation of manual audits for secrets in commit history seems a bit strong for level 1, but I'm reading it as a best practice rather than a requirement.

baseline/OSPS-BR.yaml
trumant: That makes a lot of sense. Will make these changes shortly

trumant:
> The recommendation of manual audits for secrets in commit history seems a bit strong for level 1, but I'm reading it as a best practice rather than a requirement.

My intention there was to point out that pre-commit checks are helpful but insufficient and that adding scheduled, automated checks via GHA and likely using the same tool used in pre-commit (gitleaks or similar)

If you read "manual" as implied we should probably copy edit a bit more.

evankanderson:
> > The recommendation of manual audits for secrets in commit history seems a bit strong for level 1, but I'm reading it as a best practice rather than a requirement.
> 
> My intention there was to point out that pre-commit checks are helpful but insufficient and that adding scheduled, automated checks via GHA and likely using the same tool used in pre-commit (gitleaks or similar)
> 
> If you read "manual" as implied we should probably copy edit a bit more.

"Periodically audit" suggested a process which was at least partially manual.  (Also, what happens if the audit flags a credential?  There isn't explicit guidance here, and I can imagine folks trying to expunge content from history, which would be... not great.)

I guess my main concern was requiring an audit process and answering the "leaked credentials" question seem like a higher bar than the rest of the level 1 criteria (and harder to measure).

baseline/OSPS-BR.yaml
```diff
@@ -465,3 +465,36 @@ controls:
...
+          - Maturity Level 1
+        recommendation: |
+          Configure .gitignore or equivalent to exclude files that may contain sensitive information. Use pre-commit hooks and automated scanning tools to detect and prevent the inclusion of sensitive data in commits.
+      - id: OSPS-BR-07.02
+        text: |
+          The project MUST define a policy for managing secrets and credentials used by the project. The policy should include guidelines for storing, accessing, and rotating secrets and credentials.
+        applicability:
+          - Maturity Level 3
+        recommendation: |
+          Document how secrets and credentials are managed and used within the project. This should include details on how secrets are stored (e.g., using a secrets management tool), how access is controlled, and how secrets are rotated or updated. Ensure that sensitive information is not hard-coded in the source code or stored in version control systems.
```
evankanderson, line 500: Thinking about the removal of audit guidance from BR-07.01 (which I agree with), I wonder whether we should say something here about "credentials where are suspected to have been compromised or leaked should be rotated promptly".

That may be too much detail, but I'm playing back the previous conversations, and I'm wondering if I lost some good in an attempt to broaden accessibility.

PR #375, opened 2025-08-14
Increase precision in definitions of common words

### LINKED ISSUE #340: Define what baseline applies to
The Security Baseline aims to measure and recommend controls for open source projects.

Unfortunately, we haven't defined what a "project" is anywhere, and we often conflate projects with source code repositories or specific released artifacts.  Some projects consist of multiple repositories (e.g. projects with a separate repository for each plugin, or projects that separate website, automation), or produce multiple packages possibly in different distribution channels.

We need a definition of project which allows producers and consumers to understand what is in and out of scope of a particular assessment of baseline compliance, as well as to enable discovery of baseline maturity given a particular resource.

funnelfiasco:
@eddie-knight and I worked on improving our wording. This helps us avoid ambiguity and overloaded terms. Note that this PR is only for updating the definitions. We will need to make follow-up PRs to update the usage once this is merged.

The idea here was to make sure we are clear about human and non-human actors when necessary, and to be more clear when we mean "project" (which has too many definitions, none of which make the Project Management Institute happy 😄), "repository", etc.

We are not aiming to have the perfect definition, just a clear definition.

Fixes #340

### DIFF
--- a/baseline/lexicon.yaml
+++ b/baseline/lexicon.yaml
@@ -8,6 +8,9 @@
 # streamline automation in the event that other
 # words or phrases should link to the term.
 #
+- term: Administrator
+  definition: |
+    Any human who can modify settings on the target resource.
 - term: Arbitrary Code
   definition: |
     Code provided by an external source that is
@@ -50,6 +53,10 @@
     checks.
   synonyms: 
     - Build and Release Pipelines
+- term: Code
+  definition: |
+    A set of deterministic instructions that a
+    computer can execute to perform specific tasks.
 - term: Change
   definition: |
     Any alteration of the project's codebase,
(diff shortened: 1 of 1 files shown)


baseline/lexicon.yaml
```diff
@@ -355,27 +348,21 @@
...
   definition: |
-    A storage location managed by a version
-    control system where the project's code,
-    documentation, and other resources are
-    stored. It tracks changes, manages
-    collaborator permissions, and includes
-    configuration options such as branch
-    protection and access controls.
+    A storage location managed by a version control
+    system where the project's code, documentation,
```
evankanderson, line 364: What does this definition mean for projects that use a wiki for documentation?  (I often find such projects more difficult to navigate, but that is a _personal_ bias that I'd aim to set aside here.)

baseline/lexicon.yaml
```diff
@@ -305,6 +294,10 @@
     - P-SSCRM 
   references: 
     - https://arxiv.org/pdf/2404.12300 
+- term: Project
+  definition: |
+    A group of people and resources that coordinate to
```
evankanderson, line 299: Do we define "resource" in the glossary?  I couldn't find it in the expected alphabetical order.

Do we expect a project to have a stable identity over time, even if the group of people and computing resources change?  It feels like a project should have the general property that "the project can enumerate the resources it contains, and the resources belonging to a project agree that they belong to that project" (i.e. a two-way handshake).  I'm not sure if the human membership has the same two-way handshake property, although there's an obvious difference that humans can participate in more than one project.

Note that my definition of project-resource association is **exclusive** -- resources belong to _at most_ one project, though projects might also have a sub-project relationship with other projects in a more loosely-defined hierarchy.

evankanderson (approved):
Sorry I didn't get to review this earlier!  It seems like an improvement, but I feel like the two-way resource-project handshake may be an important element for being able to automate collection of baseline information by consumers.

PR #396, opened 2025-09-28
refactor: AC-04.01 control for clarity

### LINKED ISSUE #395: AC-04.01 - rewrite the control text for clarity
## Existing control text

When a CI/CD task is executed with no permissions specified, the project's version control system MUST default to the lowest available permissions for all activities in the pipeline.

## Issues with the existing text

- "CI/CD task" could be more clear, AC-04.02 refers mainly to "job"
- CI/CD activities occur in CI/CD platforms/pipelines, but "version control system MUST" muddies the waters. This may have been written with GitHub Actions in mind specifically.

trumant:
This closes #395

### DIFF
--- a/baseline/OSPS-AC.yaml
+++ b/baseline/OSPS-AC.yaml
@@ -290,8 +290,8 @@ controls:
       - id: OSPS-AC-04.01
         text: |
           When a CI/CD task is executed with no permissions specified, the
-          project's version control system MUST default to the lowest available
-          permissions for all activities in the pipeline.
+          CI/CD system MUST default the task's permissions to the lowest
+          permissions granted in the pipeline.
         applicability:
           - Maturity Level 2
           - Maturity Level 3


baseline/OSPS-AC.yaml
```diff
@@ -290,8 +290,8 @@ controls:
       - id: OSPS-AC-04.01
         text: |
           When a CI/CD task is executed with no permissions specified, the
-          project's version control system MUST default to the lowest available
-          permissions for all activities in the pipeline.
+          CI/CD system MUST default the task's permissions to the lowest
+          permissions granted in the pipeline.
```
puerco, line 294: Non blocking nit:
```suggestion
          permissions required by the task.
```

puerco (approved):
Good catch!! Small nit below, but I'm fine with the change. Thanks @trumant @trumant !

baseline/OSPS-AC.yaml
trumant: I like your suggestion but I think it does change the focus of the requirement from one in which the task's permissions are not configured/unspecified and therefore the task runs in the presumably more well defined and least privilege permission context of it's (the task's) enclosing pipeline to one in which the enclosing pipeline should more proactively try to understand the permission requirements of the task and assign only those.

funnelfiasco: Are there CI/CD systems that use the highest permissions as a default? Or is the intent here to say "tasks should have read-only permissions unless explicitly granted more permissions at the task level"?

For example, if I have a github workflow that grants write at the top level of the workflow, would a job in that workflow that inherits write permission violate this control? Does the answer to that question depend on whether or not the job needs write permission?

If the answer is yes (or at least a conditional yes), maybe we're better off being more direct and saying "CI/CD tasks that require permissions beyond read must have permissions explicitly granted at every step" (except with better wording)

trumant:
https://github.com/ossf/security-baseline/commit/a65bd5d9937193a474b424d93a740d8452d5d62e provides some historical context around prior intent

funnelfiasco (approved):

PR #412, opened 2025-10-16
Add notes on tagging

funnelfiasco:
And provide links to the mailing list, that exists

Signed-off-by: Ben Cotton <ben@kusari.dev>

### DIFF
--- a/docs/maintenance.md
+++ b/docs/maintenance.md
@@ -10,11 +10,15 @@ At least two project maintainers must review and approve these changes.
 * Substantive changes to controls, including changes to text that alters the originally stated meaning, new controls proposals, or removal of controls will be documented in GitHub PR(s) and reviewed regularly by the Baseline project maintainers for inclusion in the next release.
 These changes may reflect changes to global cybersecurity regulations and frameworks or changes in norms around application/project security practices.
 Any such substantive changes must be approved by a majority of the project's maintainers.
+* Any changes to the Baseline will be reflected within the Compliance Matrix, with new requirements flagged where the Baseline Controls are appropriate.
+
+## Versions/releases
+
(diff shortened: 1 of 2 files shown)


funnelfiasco:
I will do the tagging once we've merged this, as that will be my signal that we're happy with the process that's loosely proposed

trumant (approved):

docs/maintenance.md
```diff
@@ -10,11 +10,15 @@ At least two project maintainers must review and approve these changes.
...
 Any such substantive changes must be approved by a majority of the project's maintainers.
+* Any changes to the Baseline will be reflected within the Compliance Matrix, with new requirements flagged where the Baseline Controls are appropriate.
+
+## Versions/releases
+
 * As appropriate, but at least annually, the Baseline project maintainers will evaluate the set of controls and, if necessary, publish a new version of the Baseline.
 Previous versions of the Baseline will remain available, but are stable and not subject to change, except for minor changes to fix technical or typographic errors.
-* Any changes to the Baseline will be reflected within the Compliance Matrix, with new requirements flagged where the Baseline Controls are appropriate.
+* Git tags are used as a convenience, but may not be updated if changes are made as described above.
 * Versions will follow a calendar-based identification system, using the `YYYY-MM-DD` format.
```
puerco, line 20: nit: I infer the git tag will follow the date schema? Directly or will it use the v convention to avoid starting with a digit?

funnelfiasco: > I infer the git tag will follow the date schema?

Yes, that is the intent. I will make that more clear after...

> Directly or will it use the v convention to avoid starting with a digit?

I would prefer consistency, but having it start with a `v` makes it unambiguously not a commit hash, which is valuable. Do you have a strong opinion either way? (Or anyone else?)

PR #420, opened 2025-11-14
Canonicalize reference links

evankanderson:
As noticed during Kubecon, the current generator sometimes generates text like `\[code] \[repositories]`, which GitHub (incorrectly?) renders as equivalent to the markdown `\[code]\[code] \[repositories]\[repositories]`.  Jekyll (and other markdown engines) seem to render this as equivalent to `\[code]\[repositories]` -- that is, a link from the text `code` to the reference link `repositories`.

By canonicalizing and always including the term link, we avoid this ambiguity, at the cost that the underlying markdown can get fairly noisy.  I'm willing to consider other solutions, but this seemed the most straight-forward.  It also removes the need to add additional reference links for all the variations on a term.

Sample markdown diff output is here, since we don't have a good way to render this:

https://gist.github.com/evankanderson/96e841f8c8f9004e73489843597cdb09

A few other bugs squashed along the way:

* CI/CD links used an anchor like `ci/cd-pipeline`, which does not seem to work.
* Reduced the number of times we iterate over the lexicon from N^3 to N^2.  This is probably a minor performance improvement.

### DIFF
--- a/docs/versions/2025-10-10.md
+++ b/docs/versions/2025-10-10.md
@@ -53,232 +53,294 @@ For more information on the project and to make contributions, visit the [GitHub
 
 ### Level 1
 
-**[OSPS-AC-01.01](#osps-ac-0101)**: When a [user] attempts to read or modify a [sensitive resource] in the [project]&#39;s
-authoritative [repository], the system MUST require the [user] to complete
-a [multi-factor authentication] process.
+**[OSPS-AC-01.01](#osps-ac-0101)**: 
+When a [user][User] attempts to read or modify a [sensitive resource][Sensitive Resource] in the [project][Project]&#39;s
+authoritative [repository][Repository], the system MUST require the [user][User] to complete
+a [multi-factor authentication][Multi-factor Authentication] process.
 
-**[OSPS-AC-02.01](#osps-ac-0201)**: When a new [collaborator] is added, the [version control system] MUST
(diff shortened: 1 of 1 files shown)


docs/versions/2025-10-10.md
```diff
@@ -53,232 +53,294 @@ For more information on the project and to make contributions, visit the [GitHub
 
 ### Level 1
 
-**[OSPS-AC-01.01](#osps-ac-0101)**: When a [user] attempts to read or modify a [sensitive resource] in the [project]&#39;s
-authoritative [repository], the system MUST require the [user] to complete
-a [multi-factor authentication] process.
-
-**[OSPS-AC-02.01](#osps-ac-0201)**: When a new [collaborator] is added, the [version control system] MUST
-require manual permission assignment, or restrict the [collaborator]
+**[OSPS-AC-01.01](#osps-ac-0101)**:
```
evankanderson, line 56: Why did this end up adding a newline to the output?

eddie-knight: I don't see the cause either, might be a matter of the yaml having newlines. But it's irrelevant because markdown doesn't do anything with single newlines, they're treated as a single paragraph.

evankanderson: Agreed.  I'm just curious because I'd tried to minimize extra changes when re-generating the YALM.

PR #426, opened 2025-11-18
Add multiple cybersecurity frameworks to YAML file

SecurityCRob:
Preparing to add new framework mappings by adding several new frameworks to the yaml

PR #427, opened 2025-11-18
Add UKSSCOP reference-ids and claims to OSPS-AC.yaml

SecurityCRob:
Dependent upon merge of  https://github.com/ossf/security-baseline/pull/426

AC mappings to UKSSCOP framework

### DIFF
--- a/baseline/OSPS-AC.yaml
+++ b/baseline/OSPS-AC.yaml
@@ -142,7 +142,6 @@ controls:
           - reference-id: CM-7
       - reference-id: UKSSCOP
         entries:
-          - reference-id: Claim 2.1.2
           - reference-id: Claim 2.2.2   
     assessment-requirements:
       - id: OSPS-AC-02.01
--- a/baseline/OSPS-AC.yaml
+++ b/baseline/OSPS-AC.yaml
@@ -299,7 +299,7 @@ controls:
           - reference-id: CM-7
       - reference-id: UKSSCOP
         entries:
-          - reference-id: Claim 2.1.2
+          - reference-id: Claim 2.1.1
           - reference-id: Claim 2.1.3   
           - reference-id: Claim 2.2.2   
     assessment-requirements:
--- a/baseline/OSPS-AC.yaml
+++ b/baseline/OSPS-AC.yaml
@@ -71,6 +71,11 @@ controls:
           - reference-id: IA-5
           - reference-id: 1.2e
           - reference-id: 1.2f
+      - reference-id: UKSSCOP
+        entries:
+          - reference-id: Claim 1.4.2
+          - reference-id: Claim 2.1.5   
+          - reference-id: Claim 2.2.2   
     assessment-requirements:
       - id: OSPS-AC-01.01
         text: |
@@ -135,6 +140,10 @@ controls:
           - reference-id: AC-6
           - reference-id: CM-5
           - reference-id: CM-7
+      - reference-id: UKSSCOP
+        entries:
+          - reference-id: Claim 2.1.2
+          - reference-id: Claim 2.2.2   
     assessment-requirements:
       - id: OSPS-AC-02.01
         text: |
@@ -199,6 +208,9 @@ controls:
           - reference-id: CM-3
           - reference-id: CM-3(2)
           - reference-id: CM-5
+      - reference-id: UKSSCOP
+        entries:
+          - reference-id: Claim 2.2.2
     assessment-requirements:
       - id: OSPS-AC-03.01
         text: |
@@ -286,6 +298,11 @@ controls:
           - reference-id: AC-20(1)
           - reference-id: CM-5
           - reference-id: CM-7
+      - reference-id: UKSSCOP
+        entries:
+          - reference-id: Claim 2.1.2
+          - reference-id: Claim 2.1.3   
+          - reference-id: Claim 2.2.2   
     assessment-requirements:
       - id: OSPS-AC-04.01
         text: |
--- a/baseline/OSPS-AC.yaml
+++ b/baseline/OSPS-AC.yaml
@@ -209,6 +209,7 @@ controls:
           - reference-id: CM-5
       - reference-id: UKSSCOP
         entries:
+          - reference-id: Claim 1.1.4
           - reference-id: Claim 2.2.2
     assessment-requirements:
       - id: OSPS-AC-03.01
--- a/baseline/OSPS-AC.yaml
+++ b/baseline/OSPS-AC.yaml
@@ -58,9 +58,6 @@ controls:
           - reference-id: 2.2.1
           - reference-id: 8.2.1
           - reference-id: 8.3.1
-      - reference-id: UKSSCOP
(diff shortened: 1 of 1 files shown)


baseline/OSPS-AC.yaml
```diff
@@ -71,6 +71,11 @@ controls:
           - reference-id: IA-5
           - reference-id: 1.2e
           - reference-id: 1.2f
+      - reference-id: UKSSCOP
+        entries:
+          - reference-id: Claim 1.4.2
```
funnelfiasco, line 73: 1.4.2 says "privileged users of the software", but this control is about the project infrastructure, so I don't think this is a match.

baseline/OSPS-AC.yaml
```diff
@@ -135,6 +140,10 @@ controls:
           - reference-id: AC-6
           - reference-id: CM-5
           - reference-id: CM-7
+      - reference-id: UKSSCOP
+        entries:
+          - reference-id: Claim 2.1.2
```
funnelfiasco: I don't think this fits.

baseline/OSPS-AC.yaml
```diff
@@ -199,6 +208,9 @@ controls:
           - reference-id: CM-3
           - reference-id: CM-3(2)
           - reference-id: CM-5
+      - reference-id: UKSSCOP
+        entries:
+          - reference-id: Claim 2.2.2
```
funnelfiasco, line 203: 2.2.2 is about the build system, not the code. I don't think there's an equivalent UKSSCOP claim to this control.

baseline/OSPS-AC.yaml
```diff
@@ -286,6 +298,11 @@ controls:
           - reference-id: AC-20(1)
           - reference-id: CM-5
           - reference-id: CM-7
+      - reference-id: UKSSCOP
+        entries:
+          - reference-id: Claim 2.1.2
```
funnelfiasco:
```suggestion
          - reference-id: Claim 2.1.1
```

baseline/OSPS-AC.yaml
SecurityCRob: A counter.... the section is titled "Principle 1.4 - Follow secure by design and secure by default principles throughout the development lifecycle of the software."  We are primarily focused on maintainers, who are considered privileged users of the system.  Having this control in place during a development phase is crucial to help protect the codebase from merging unintended/malicious commits by giving a higher degree of assurance that the maintainer is who they state they are.

baseline/OSPS-AC.yaml
SecurityCRob: i can see your objection.  At best, AC02 only partially covers "Users of the build environment are required to authenticate on a regular basis."  Since we don't have anything closer, I'll remove it.

baseline/OSPS-AC.yaml
SecurityCRob: AC03 is about preventing changes to the main branch of software, which lives in the VCS.  Branch Protection is part of the git repo's config settings.  I feel AC03 is a good match for Claim 2.2.2.  AC03 does not cover any other config settings that could be considered "changes in the build environment"

baseline/OSPS-AC.yaml
```diff
@@ -71,6 +71,11 @@ controls:
           - reference-id: IA-5
           - reference-id: 1.2e
           - reference-id: 1.2f
+      - reference-id: UKSSCOP
+        entries:
+          - reference-id: Claim 1.4.2
+          - reference-id: Claim 2.1.5   
+          - reference-id: Claim 2.2.2
```
evankanderson, line 75: I think 2.2.2 reads:

> Only authorised personnel can make changes to the build environment.

I'm not sure that's a clean mapping to requiring MFA for these.  I think it might better map to AC-02.01 (which I see your also did).

baseline/OSPS-AC.yaml
evankanderson: > The ‘build environment’ refers to the environment where software is compiled, built and packaged ready for release.

This is about where the software is _stored_.  This feels closest to 1.1.4:

> Items requiring configuration control are identified and version control is used.

Though this is more specific about version control than that document.

baseline/OSPS-AC.yaml
```diff
@@ -286,6 +297,11 @@ controls:
           - reference-id: AC-20(1)
           - reference-id: CM-5
           - reference-id: CM-7
+      - reference-id: UKSSCOP
+        entries:
+          - reference-id: Claim 2.1.1
+          - reference-id: Claim 2.1.3
```
evankanderson, line 290: I'm not sure that this item is intended to talk about project contributors' permissions _on_ the build pipeline.  (From the recommendation: "limiting the permissions granted to steps within the CI/CD pipelines")

2.1.3 seems like it is closer to a build-system equivalent of AC-02.01, though I read it as simply saying "everyone gets their own login on e.g. Jenkins -- don't use a shared account".

baseline/OSPS-AC.yaml
SecurityCRob: The maintainers are the most privileged users of all.  With our focus on the build-phrase (predominantly) for this control, this does not fulfill completely the UK's requirement, but it supports it as development/build are parts of "throughout the development lifecycle".  Not everyone one of these controls will be a 1:1 direct mapping given the multitude of audiences all of these frameworks were developed to serve; the mappings provided help explain where we feel there are supporting connections.  As I went through these I tagged things that were COMPLETE coverage, PARTIAL, or SUPPORTS.  I suppose we could add more detail, if we want to refactor how we're doing this, to better show those three states, we certainly could, but for now if any of these three states were applicable I added in a map entry for it.

baseline/OSPS-AC.yaml
SecurityCRob: good catch, i'll add that, ty

baseline/OSPS-AC.yaml
SecurityCRob: UK Principle 2.1 is about "Protect the build environment against unauthorized access". AC04 is about "The project's permissions in CI/CD pipelines MUST follow the principle of least privilege."  I feel these two Claims are strongly aligned with our AC04

funnelfiasco (dismissed):

evankanderson (commented):
This file also has existing UKSSCOP mappings of the form:

```yaml
      - reference-id: UKSSCOP
        entries:
          - reference-id: 2.1
```

See e.g. lines 61-63.  Do we need these additional mappings, and should they be merged?

SecurityCRob:
> This file also has existing UKSSCOP mappings of the form:
> 
> ```yaml
>       - reference-id: UKSSCOP
>         entries:
>           - reference-id: 2.1
> ```
> 
> See e.g. lines 61-63. Do we need these additional mappings, and should they be merged?

These need to be removed, I'll go take care of that.  thanks!

funnelfiasco (approved):

eddie-knight (approved):

PR #428, opened 2025-11-18
Add UKSSCOP reference-ids and claims to OSPS-BR.yaml 

SecurityCRob:
Added UKSSCOP reference IDs and claims to multiple sections.

Dependent upon merge of #426

BR mappings to UKSSCOP framework

### DIFF
--- a/baseline/OSPS-BR.yaml
+++ b/baseline/OSPS-BR.yaml
@@ -57,6 +57,10 @@ controls:
           - reference-id: CM-5
           - reference-id: CM-7
           - reference-id: SI-7
+      - reference-id: UKSSCOP
+        entries:
+          - reference-id: Claim 2.1.2
+          - reference-id: Claim 2.2.2          
     assessment-requirements:
       - id: OSPS-BR-01.01
         text: |
@@ -127,6 +131,10 @@ controls:
           - reference-id: SA-15
           - reference-id: SI-7
           - reference-id: SR-4
+      - reference-id: UKSSCOP
+        entries:
+          - reference-id: Claim 1.1.4
+          - reference-id: Claim 3.1.1  
     assessment-requirements:
       - id: OSPS-BR-02.01
         text: |
@@ -204,6 +212,10 @@ controls:
         entries:
           - reference-id: AC-4
           - reference-id: AC-4(21)
+      - reference-id: UKSSCOP
+        entries:
+          - reference-id: Claim 2.1.2
+          - reference-id: Claim 3.1.2  
     assessment-requirements:
       - id: OSPS-BR-03.01
         text: |
@@ -307,6 +319,11 @@ controls:
           - reference-id: MA-8
           - reference-id: SI-4
           - reference-id: SI-5
+      - reference-id: UKSSCOP
+        entries:
+          - reference-id: Claim 1.1.4
+          - reference-id: Claim 2.2.3
+          - reference-id: Claim 3.1.1
     assessment-requirements:
       - id: OSPS-BR-04.01
         text: |
@@ -387,6 +404,10 @@ controls:
           - reference-id: RA-5
           - reference-id: SA-15
           - reference-id: SR-3
+      - reference-id: UKSSCOP
+        entries:
+          - reference-id: Claim 2.1.2
+          - reference-id: Claim 3.1.1  
     assessment-requirements:
       - id: OSPS-BR-05.01
         text: |
@@ -451,6 +472,10 @@ controls:
           - reference-id: SA-15
           - reference-id: SI-7
           - reference-id: SI-7(14)
+      - reference-id: UKSSCOP
+        entries:
+          - reference-id: Claim 1.2.2
+          - reference-id: Claim 3.1.1  
     assessment-requirements:
       - id: OSPS-BR-06.01
         text: |
@@ -483,6 +508,9 @@ controls:
           - reference-id: PW.1.2
           - reference-id: PW.1.3
           - reference-id: PW.5.1
+      - reference-id: UKSSCOP
+        entries:
+          - reference-id: Claim 3.1.2
     assessment-requirements:
       - id: OSPS-BR-07.01
         text: |
--- a/baseline/OSPS-BR.yaml
+++ b/baseline/OSPS-BR.yaml
@@ -509,7 +509,8 @@ controls:
           - reference-id: PW.5.1
       - reference-id: UKSSCOP
         entries:
-          - reference-id: Claim 3.1.2
+          - reference-id: Claim 1.4.3
+          - reference-id: Claim 1.4.5
     assessment-requirements:
       - id: OSPS-BR-07.01
         text: |
--- a/baseline/OSPS-BR.yaml
+++ b/baseline/OSPS-BR.yaml
@@ -135,6 +135,7 @@ controls:
         entries:
           - reference-id: Claim 1.1.4
           - reference-id: Claim 3.1.1  
+          - reference-id: Claim 3.4.2 
     assessment-requirements:
       - id: OSPS-BR-02.01
         text: |
--- a/baseline/OSPS-BR.yaml
+++ b/baseline/OSPS-BR.yaml
@@ -122,9 +122,6 @@ controls:
       - reference-id: PCIDSS
         entries:
           - reference-id: 6.4.3
-      - reference-id: UKSSCOP
-        entries:
-          - reference-id: 3.1
       - reference-id: 800-161
         entries:
           - reference-id: IA-4
@@ -206,9 +203,6 @@ controls:
           - reference-id: 4.2.2
           - reference-id: 6.4.1
           - reference-id: 8.3.2
-      - reference-id: UKSSCOP
-        entries:
-          - reference-id: 3.1
       - reference-id: 800-161
         entries:
           - reference-id: AC-4
@@ -304,10 +298,6 @@ controls:
           - reference-id: 6.5.1
           - reference-id: 6.5.2
           - reference-id: 10.2.2
-      - reference-id: UKSSCOP
-        entries:
-          - reference-id: 3.1
-          - reference-id: 3.5
       - reference-id: 800-161
         entries:
           - reference-id: AU-2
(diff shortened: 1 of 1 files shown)


baseline/OSPS-BR.yaml
```diff
@@ -57,6 +57,10 @@ controls:
           - reference-id: CM-5
           - reference-id: CM-7
           - reference-id: SI-7
+      - reference-id: UKSSCOP
+        entries:
```
funnelfiasco, line 61: This feels a little tenuous, especially 2.1.2. I can certainly see an argument though, so I'll just leave this comment for now and see if anyone else weighs in. If not, I think we proceed as-is.

baseline/OSPS-BR.yaml
```diff
@@ -127,6 +131,10 @@ controls:
           - reference-id: SA-15
           - reference-id: SI-7
           - reference-id: SR-4
+      - reference-id: UKSSCOP
+        entries:
+          - reference-id: Claim 1.1.4
+          - reference-id: Claim 3.1.1
```
funnelfiasco, line 134: BR-02 is a part of Claim 3.1.1, but BR-06 is a better fit. I'm inclined to drop it here.

baseline/OSPS-BR.yaml
```diff
@@ -204,6 +212,10 @@ controls:
         entries:
           - reference-id: AC-4
           - reference-id: AC-4(21)
+      - reference-id: UKSSCOP
+        entries:
+          - reference-id: Claim 2.1.2
```
funnelfiasco: I don't see BR-03 as being relevant to 2.1.2. 2.1.2 implies access controls, which this control doesn't touch.

baseline/OSPS-BR.yaml
```diff
@@ -307,6 +319,11 @@ controls:
           - reference-id: MA-8
           - reference-id: SI-4
           - reference-id: SI-5
+      - reference-id: UKSSCOP
+        entries:
+          - reference-id: Claim 1.1.4
```
funnelfiasco, line 314: 1.1.4 is kind of tangential here, although `git log` would technically meet the control, so I guess I'll allow it :-)

baseline/OSPS-BR.yaml
```diff
@@ -307,6 +319,11 @@ controls:
...
+          - reference-id: Claim 2.2.3
```
funnelfiasco, line 315: 2.2.3 is about build system logs, whereas this is the software changelog. I think we drop this reference.

baseline/OSPS-BR.yaml
```diff
@@ -307,6 +319,11 @@ controls:
...
+          - reference-id: Claim 3.1.1
```
funnelfiasco, line 316: I don't see the connection here. I read 3.1.1 as being about cryptographic hash validation.

baseline/OSPS-BR.yaml
```diff
@@ -387,6 +404,10 @@ controls:
           - reference-id: RA-5
           - reference-id: SA-15
           - reference-id: SR-3
+      - reference-id: UKSSCOP
+        entries:
+          - reference-id: Claim 2.1.2
+          - reference-id: Claim 3.1.1
```
funnelfiasco:
```suggestion
          - reference-id: Claim 1.2.2.1.1
```

I think the intent here more closely matches "Integrity of third-party components and updates is validated"

baseline/OSPS-BR.yaml
```diff
@@ -451,6 +472,10 @@ controls:
           - reference-id: SA-15
           - reference-id: SI-7
           - reference-id: SI-7(14)
+      - reference-id: UKSSCOP
+        entries:
+          - reference-id: Claim 1.2.2
```
funnelfiasco, line 460: This control is more about output than input. 1.2.2 is about input, and 3.1.1 is output, so we should drop 1.2.2 here imo.

baseline/OSPS-BR.yaml
```diff
@@ -483,6 +508,9 @@ controls:
           - reference-id: PW.1.2
           - reference-id: PW.1.3
           - reference-id: PW.5.1
+      - reference-id: UKSSCOP
+        entries:
+          - reference-id: Claim 3.1.2
```
funnelfiasco:
```suggestion
          - reference-id: Claim 1.4.3
          - reference-id: Claim 1.4.5
```

baseline/OSPS-BR.yaml
```diff
@@ -127,6 +131,10 @@ controls:
           - reference-id: SA-15
           - reference-id: SI-7
           - reference-id: SR-4
+      - reference-id: UKSSCOP
+        entries:
+          - reference-id: Claim 1.1.4
```
funnelfiasco, line 133: I don't think 1.1.4 is a great fit here.

baseline/OSPS-BR.yaml
SecurityCRob: Claim 2.1.2 "Users of the build environment are required to authenticate on a regular basis."  Is Partial coverage at best.  A means of ensuring that actors & inputs are trusted to to periodically authenticate them.

baseline/OSPS-BR.yaml
SecurityCRob: Claim 1.1.4 "Items requiring configuration control are identified and version control is used."   BR02 "All releases and released software assets MUST be assigned a unique version identifier for each release intended to be used by users."  We are predominantly focused on the application software, but this is equally applicable to "config as code" as well, that all needs clear version ids

baseline/OSPS-BR.yaml
SecurityCRob: I like BOTH!

baseline/OSPS-BR.yaml
funnelfiasco: Okay, you sold me.

baseline/OSPS-BR.yaml
SecurityCRob: thanks!

baseline/OSPS-BR.yaml
evankanderson: I think this may be a distinction between commercial and OSS software -- in OSS, we expect that attackers can do a lot more in the build system than you would expect for commercial software.  So this hardening is roughly because OSS doesn't enforce 2.1.6 "Users with access to the build environment are regularly reviewed to ensure they still have a legitimate need".

baseline/OSPS-BR.yaml
evankanderson: The BR-02.01 requirement has always been a little confusing for me (mostly due to a missing counterexample / bad practice).  BR-02.02 seems a bit more understandable.  Artifact version identification feels like something obvious enough that maybe it's not listed.  On the other hand, I think 3.4.2 could also apply here -- "Affected customers are informed" (e.g. you need distinct version numbers to say "version X is affected, version Y is not").

baseline/OSPS-BR.yaml
evankanderson: Yeah, I'd say this is pretty clearly 3.1.2

baseline/OSPS-BR.yaml
evankanderson: Your suggestion didn't come through clearly "1.2.2.1.1" maybe means "1.2.1"?

evankanderson: 1.2.5 seems like it should apply as well.

baseline/OSPS-BR.yaml
evankanderson: I think 2.1.4 ("Credentials are securely managed and stored") would apply here as well.

baseline/OSPS-BR.yaml
SecurityCRob: agreed

baseline/OSPS-BR.yaml
SecurityCRob: agreed

baseline/OSPS-BR.yaml
SecurityCRob: removed 2.1.2 and 3.1.1 added 1.2.1 and 1.2.5

baseline/OSPS-BR.yaml
SecurityCRob: BR-06 enables 1.2.2 and 3.1.1

baseline/OSPS-BR.yaml
SecurityCRob: logging changes assists consumers in verifying the integrity of the software they are deploying and understanding the full impact of those changes to their systems

baseline/OSPS-BR.yaml
SecurityCRob: agreed.  any adjustments desired?

baseline/OSPS-BR.yaml
SecurityCRob: i'll add in 3.4.2.  thanks for the observation

SecurityCRob (commented):
updated or responded based on feedback

funnelfiasco (dismissed):

baseline/OSPS-BR.yaml
evankanderson: I might consider this 2.1.1 ("Roles are defined that specify the data and functionality that each role is allowed to access") in combination with 2.2.2.  I'd generally say that these mappings are pretty weak, and I might be inclined to simply remove them and say that OSS has a different approach from UKSSCOP.

baseline/OSPS-BR.yaml
evankanderson: I think we said we'd drop 2.2.3, but it's still here.

baseline/OSPS-BR.yaml
evankanderson: I agree with Ben -- the claim is that "the _integrity_ of software (including updates)..." (my emphasis).  I don't think a descriptive log of the changes passes that test; I'd expect something involving hashes and signatures.  (e.g. BR-06)

baseline/OSPS-BR.yaml
evankanderson: I just noticed that we have an entry for "UKSSCOP" to 3.1 on line 446.

In fact, we already seem to have mappings for UKSSCOP which disagree with the more fine-grained ones we are making now.

evankanderson (commented):
I just noticed that we have entries for UKSSCOP (e.g. lines 295-298, 444-446).  Those reference the "principles" of UKSSCOP, rather than the "claims", and don't seem to align with the current mapping.

funnelfiasco (approved):

eddie-knight (approved):

PR #431, opened 2025-11-18
Add UKSSCOP reference-ids and claims to OSPS-QA.yaml

SecurityCRob:
Added new reference IDs under UKSSCOP and updated existing ones.

Dependent upon merge of #426

QA mappings to UKSSCOP framework

### DIFF
--- a/baseline/OSPS-QA.yaml
+++ b/baseline/OSPS-QA.yaml
@@ -173,7 +173,6 @@ controls:
           - reference-id: SR-4                
       - reference-id: UKSSCOP
         entries:
-          - reference-id: Claim 1.1.4
           - reference-id: Claim 1.2.1
           - reference-id: Claim 1.2.2
           - reference-id: Claim 3.1.1
--- a/baseline/OSPS-QA.yaml
+++ b/baseline/OSPS-QA.yaml
@@ -69,6 +69,9 @@ controls:
           - reference-id: RA-5
           - reference-id: SA-11
           - reference-id: SA-15          
+      - reference-id: UKSSCOP
+        entries:
+          - reference-id: Claim 2.2.3
     assessment-requirements:
       - id: OSPS-QA-01.01
         text: |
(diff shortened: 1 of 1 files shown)


baseline/OSPS-QA.yaml
```diff
@@ -69,6 +69,9 @@ controls:
           - reference-id: RA-5
           - reference-id: SA-11
           - reference-id: SA-15          
+      - reference-id: UKSSCOP
+        entries:
+          - reference-id: Claim 2.2.3
```
funnelfiasco, line 74: This is a little tenuous, IMO, as the baseline control is about source code and the Claim is about build logs, but I can see some overlap. I don't object to keeping it, but flagging for others as they review.

baseline/OSPS-QA.yaml
```diff
@@ -171,6 +171,12 @@ controls:
           - reference-id: SA-15
           - reference-id: SR-3
           - reference-id: SR-4                
+      - reference-id: UKSSCOP
+        entries:
+          - reference-id: Claim 1.1.4
+          - reference-id: Claim 1.2.1
+          - reference-id: Claim 1.2.2
+          - reference-id: Claim 3.1.1
```
funnelfiasco, line 178: I don't think these two fit as well as 1.2.1.

SecurityCRob: ack on 1.1.4

baseline/OSPS-QA.yaml
SecurityCRob: as I stated in https://github.com/ossf/security-baseline/pull/427/files#r2565105191, when doing these mappings, I'd sort them into COMPLETE (coverage), PARTIAL (Coverage), SUPPORTS (as I feel the case is here), or N/A NONE (does not apply).  If any of the 1st three states were triggered as I was doing my review, I'd "map" it.  If we want to, we all could go back and refactor all of our mappings to show how strong the connection is to the framework under consideration.  I'm unsure how that might complicate things, but if the team is interested in pursuing that, I certainly would be glad to contribute.

funnelfiasco (approved):

eddie-knight (approved):

PR #440, opened 2025-11-25
Clarify distribution security requirements

evankanderson:
As discussed in the 2025-11-25 meeting, correct the recommendation for OSPS-BR-03.02.

### DIFF
--- a/baseline/OSPS-BR.yaml
+++ b/baseline/OSPS-BR.yaml
@@ -234,7 +234,7 @@ controls:
         text: |
           When the project lists a URI as an official distribution channel,
           that channel MUST be protected from adversary-in-the-middle
-          attacks.
+          attacks using cryptographically authenticated channels.
         applicability:
           - Maturity Level 1
           - Maturity Level 2


puerco (dismissed):

baseline/OSPS-BR.yaml
```diff
@@ -221,15 +221,18 @@ controls:
       - id: OSPS-BR-03.02
         text: |
           When the project lists a URI as an official distribution channel,
-          that URI MUST be exclusively delivered using encrypted channels.
+          that distribution MUST be protected from adversary-in-the-middle
```
funnelfiasco:
```suggestion
          that channel MUST be protected from adversary-in-the-middle
```
Is this better? I'm not sure. Even with context, when my brain saw "that distribution", I thought "Linux distro".

evankanderson: I'm happy to use that language.

baseline/OSPS-BR.yaml
```diff
@@ -221,15 +221,18 @@ controls:
       - id: OSPS-BR-03.02
         text: |
           When the project lists a URI as an official distribution channel,
-          that URI MUST be exclusively delivered using encrypted channels.
+          that channel MUST be protected from adversary-in-the-middle
+          attacks.
```
eddie-knight: 1. What is the functional difference between the old and new text?
2. If there is value added by this, why isn't `BR-03.01` also changed to get the same value?

evankanderson: (Sorry, put this in a box for a while since we didn't have a meeting.)

1. The old text required that the content was delivered using an encrypted channel, but the actual intent was that the content was delivered in a signed or otherwise authenticated manner.  This could include delivery via HTTP or FTP with e.g. a PGP signature which was checked against an existing (local) key, as many Linux distributions were originally designed to do.

2. BR-03.01's "official project channels" doesn't have existing prior art (RPM / DEB / etc ecosystems) which was designed for secure distribution prior to the widespread adoption of TLS.  While HTTPS is a simple way to meet BR-03.02's requirements, there are _existing_ systems which meet the requirements without the use of HTTPS which we're not looking to exclude.  BR-03.01 does not have a comparable existing ecosystem that I'm aware of.

evankanderson: Suggested by @david-a-wheeler :

```suggestion
          that channel MUST be protected from adversary-in-the-middle
          attacks using cryptographically authenticated channels.
```

funnelfiasco (approved):

trumant (approved):

PR #452, opened 2025-12-17
Update OSCAL part names and subclass reqs into items

### LINKED ISSUE #451: OSCAL validation checks failing
Our OSCAL validation has apparently been failing for some time. See, e.g., https://github.com/ossf/security-baseline/actions/runs/20123940750/job/57749652527

However, we only notice it now because #450 fixes a bug in the action that caused the errors to not be reflected in the exit code.

puerco:
This PR should fix the OSCAL validation failures from the new validation performed by the oscal-club action. The oscal generator now uses the proper part depending on the nesting level. The resulting oscal code now subclasses each control requirement (eg `OSPS-AC-03.01`) into items in the control definition.

Fixes #451 
Supersedes #450 

Signed-off-by: Adolfo García Veytia (Puerco) <puerco@carabiner.dev>

### DIFF
--- a/cmd/pkg/baseline/generator_oscal.go
+++ b/cmd/pkg/baseline/generator_oscal.go
@@ -51,50 +51,65 @@ func (g *Generator) ExportOSCAL(b *types.Baseline, w io.Writer) error {
 			Class:    "OSPS",
 			Controls: nil,
 			ID:       b.ControlFamilyIDs[family.Title],
-			Title:    family.Description,
+			Parts: &[]oscal.Part{
+				{
+					Name:  "overview",
+					Prose: family.Description,
+				},
+			},
+			Title: family.Title,
 		}
 
 		controls := []oscal.Control{}
 		for _, control := range family.Controls {
-			parts := []oscal.Part{}
-			for _, ar := range control.AssessmentRequirements {
-				parts = append(parts, oscal.Part{
-					Class: control.Id,
-					ID:    ar.Id,
-					Name:  ar.Id,
-					Ns:    "",
-					Parts: &[]oscal.Part{
-						{
-							ID:    ar.Id + ".R",
-							Name:  "recommendation",
-							Ns:    OpenSSFNS,
-							Prose: ar.Recommendation,
-							Links: &[]oscal.Link{
-								{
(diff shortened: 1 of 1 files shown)


.github/workflows/oscal-check.yaml
```diff
@@ -50,7 +50,7 @@ jobs:
           ./go-oscal validate --input-file oscal-sample.json 
 
       - name: Validate Catalog with oscal-club/NIST CLI
-        uses: oscal-club/oscal-cli-action@1a210b84bc1fd6adf15c9cf0d46a51d15a3d8301 # v2.0.1
+        uses: oscal-club/oscal-cli-action@b5b0c80a1a158797bea4475d13d12c494b56019b # v2.0.2
```
xee5ch, line 53: Just a heads up new action version dropped today, v2.1.0, which defaults to the newest CLI version from metaschema-framework fork.

Also those maintainers think there may be issues with the regex and filed bugs when they saw #450 and we chatted, but still green is green, you might want to get ahead of the dependabot bump in a day or two as you review this.

```suggestion
        uses: oscal-club/oscal-cli-action@6a8b6368885714e46fa1d4a65a6ce665902e08b7 # v2.1.0
```

.github/workflows/oscal-check.yaml
```diff
@@ -50,7 +50,7 @@ jobs:
           ./go-oscal validate --input-file oscal-sample.json 
 
       - name: Validate Catalog with oscal-club/NIST CLI
```
xee5ch, line 52: Minor, non-blocking: the action allows you to use either the legacy NIST `oscal-cli` or the fork maintained by the metaschema-framework, but it defaults to the latter. You might want to consider overriding the default and using the NIST, but as the m-f maintainers remind people there are important bug fixes in the latter, so I would rename this to be less confusing.

```suggestion
      - name: Validate Catalog with oscal-club/oscal-cli-action
```

That said, you all maintain it, so maybe you know to ignore the name.

SecurityCRob (approved):

.github/workflows/oscal-check.yaml
puerco: Ah thanks for the heads up @xee5ch , I'll test it. Perhaps we can even run both?

PR #459, opened 2026-01-13
Add BSI TR-03185-2 to frameworks.yaml

SecurityCRob:
Added BSI TR-03185-2 guidelines for secure software lifecycle.

PR #460, opened 2026-01-13
Add BSI TR-03185-2 mapping reference

SecurityCRob:
Added mapping reference for BSI TR-03185-2 including title, version, URL, and description.

### DIFF
--- a/baseline/metadata.yaml
+++ b/baseline/metadata.yaml
@@ -18,6 +18,16 @@ applicability-categories: #TODO: Update all applicability levels to use these ID
     description: for any code project that has a large number of consistent users
 
 mapping-references:
+   - id: BSI-TR-03185-2 
+    title: BSI TR-03185-2 Secure Software Lifecycle for Open Source Software
+    version: "1.1.0"
+    url: https://www.bsi.bund.de/SharedDocs/Downloads/EN/BSI/Publications/TechGuidelines/TR03185/BSI-TR-03185-2.pdf
+    description: |
+      Secure software is essential for the use of IT products in governments, 
+      businesses and societies. The German Federal Office for Information 
+      Security (BSI) appeals to all relevant stakeholders to consider information
+      security from the outset and to ease estimating a software component’s 
+      security posture as far as possible in order to enable its secure use.
   - id: BPB
     title: OpenSSF Best Practices Badge
     version: ""


SecurityCRob:
depends on:

related to:

baseline/metadata.yaml
```diff
@@ -18,6 +18,16 @@ applicability-categories: #TODO: Update all applicability levels to use these ID
...
+   - id: BSI-TR-03185-2 
+    title: BSI TR-03185-2 Secure Software Lifecycle for Open Source Software
+    version: "1.1.0"
+    url: https://www.bsi.bund.de/SharedDocs/Downloads/EN/BSI/Publications/TechGuidelines/TR03185/BSI-TR-03185-2.pdf
+    description: |
+      Secure software is essential for the use of IT products in governments, 
+      businesses and societies. The German Federal Office for Information 
+      Security (BSI) appeals to all relevant stakeholders to consider information
+      security from the outset and to ease estimating a software component’s 
+      security posture as far as possible in order to enable its secure use.
```
evankanderson, line 30: This looks to be the first paragraph from the regulation, which is fine.

In the larger context of these mappings, BSI TR-03185-2 is specifically focused on _open source_ projects, and BSI TR-03185-1 provides equivalent guidance for proprietary software.  This is in contrast to  unlike the UK SSCOP, which does not differentiate the two, and produced more difficulty in the mappings.

evankanderson (approved):
(I don't see changes needed, but I wanted to point out a relation with earlier mapping challenges.)

evankanderson:
One interesting note which I only realized when I was almost *done* with these reviews is that the BSI TR-03185-2 guidance is based on the 2025-02-25 version of baseline, and we have added and changed controls for the October release and the in-development release (these mappings should be applied to the in-development version, not the "current version").

baseline/metadata.yaml
SecurityCRob: i agree with your assessment.  Any suggested action?

evankanderson: Mostly, I was pointing this out as a guideline for future framework mappings.  It *might* also be worth an aside somewhere about differences between proprietary and open source software security (e.g. security by obscurity _really_ doesn't exist, things like build pipelines are generally much more exposed, and it's much easier for malicious actors to propose changes / fork code).

funnelfiasco (approved):

PR #461, opened 2026-01-13
Add BSI-TR-03185-2 reference IDs to OSPS-AC.yaml

SecurityCRob:
adding mappings for BSI

### DIFF
--- a/baseline/OSPS-AC.yaml
+++ b/baseline/OSPS-AC.yaml
@@ -143,7 +143,7 @@ controls:
       - reference-id: BSI-TR-03185-2
         entries:
           - reference-id: GV.02      
-      assessment-requirements:
+    assessment-requirements:
       - id: OSPS-AC-02.01
         text: |
           When a new collaborator is added, the version control system MUST
--- a/baseline/OSPS-AC.yaml
+++ b/baseline/OSPS-AC.yaml
@@ -140,7 +140,10 @@ controls:
       - reference-id: UKSSCOP
         entries:
           - reference-id: Claim 2.2.2   
-    assessment-requirements:
+      - reference-id: BSI-TR-03185-2
+        entries:
+          - reference-id: GV.02      
+      assessment-requirements:
       - id: OSPS-AC-02.01
         text: |
           When a new collaborator is added, the version control system MUST
@@ -206,6 +209,7 @@ controls:
           - reference-id: Claim 2.2.2
       - reference-id: BSI-TR-03185-2
         entries:
+          - reference-id: GV.02      
           - reference-id: QA.06      
     assessment-requirements:
       - id: OSPS-AC-03.01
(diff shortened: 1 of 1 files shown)


baseline/OSPS-AC.yaml
```diff
@@ -73,6 +73,9 @@ controls:
           - reference-id: Claim 1.4.2
           - reference-id: Claim 2.1.5   
           - reference-id: Claim 2.2.2   
+      - reference-id: BSI-TR-03185-2
+        entries:
+          - reference-id: GV.01
```
eddie-knight: BSI GV.01 states:

> Information on how to contribute to the project MUST be documented.
Information about the expected quality of contributions SHOULD be
given.

This documentation requirement does not appear to be relevant to OSPS-AC-01, which pertains to mandatory enforcement of multi-factor authentication.

baseline/OSPS-AC.yaml
```diff
@@ -201,6 +204,9 @@ controls:
         entries:
           - reference-id: Claim 1.1.4
           - reference-id: Claim 2.2.2
+      - reference-id: BSI-TR-03185-2
+        entries:
+          - reference-id: QA.06
```
eddie-knight, line 209: BSI QA-06 states:

> All changes to the source code SHOULD be peer-reviewed.

This seems more similar to OSPS-QA-07.01: "When a commit is made to the primary branch, the project's version control system MUST require at least one non-author human approval of the changes before merging."

I can see a loose mapping to branch protection, but it's pretty weak compared to QA-07. Do we want to keep multiple outgoing mappings?

eddie-knight (dismissed):

baseline/OSPS-AC.yaml
SecurityCRob: thanks for catching this, i updated to gv.02 -> ac-01

baseline/OSPS-AC.yaml
SecurityCRob: This mapping came from BSI (BSI qa.06 -> OSPS AC.03).  I'm fine also adding QA-07 as you point out it does appear relevant too.

evankanderson (commented):
I'm wondering if GV.02 ("The project's repository, websites and sensitive data MUST be protected against unauthorized actions") maps more broadly than specified by the PDF; it seems like it would also map to AC-02 and possibly AC-03 here, along with BR-07 and BR-01 in your other PR.

I also agree with Eddie's comment on QA-07 being a better stronger mapping for the BSI's QA.06, but that's in a different file, so I'll comment there.

funnelfiasco (dismissed):

funnelfiasco (dismissed):

funnelfiasco (approved):

eddie-knight (approved):

PR #462, opened 2026-01-13
Add BSI-TR-03185-2 reference IDs to OSPS-BR.yaml

SecurityCRob:
adding br mappings for bsi

PR #463, opened 2026-01-13
Add BSI-TR-03185-2 reference IDs to DO metadata

SecurityCRob:
adding awesomeness via bsi mappings

### DIFF
--- a/baseline/OSPS-DO.yaml
+++ b/baseline/OSPS-DO.yaml
@@ -67,6 +67,9 @@ controls:
           - reference-id: PL-2
           - reference-id: PL-8
           - reference-id: SA-15
+      - reference-id: BSI-TR-03185-2
+        entries:
+          - reference-id: BR.01          
     assessment-requirements:
       - id: OSPS-DO-01.01
         text: |
@@ -138,6 +141,9 @@ controls:
           - reference-id: IR-6
           - reference-id: SI-4
           - reference-id: SI-5          
+      - reference-id: BSI-TR-03185-2
+        entries:
+          - reference-id: QA.03
     assessment-requirements:
       - id: OSPS-DO-02.01
         text: |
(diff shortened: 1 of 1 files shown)


SecurityCRob:
depends on:

related to:

baseline/OSPS-DO.yaml
```diff
@@ -67,6 +67,9 @@ controls:
           - reference-id: PL-2
           - reference-id: PL-8
           - reference-id: SA-15
+      - reference-id: BSI-TR-03185-2
+        entries:
+          - reference-id: BR.01
```
evankanderson, line 72: While I agree that BSI TR-03185-2 lists this as a requirement that induces BR.01, I'm not sold that this is a solid mapping.  (Maybe we need a new requirement that the software includes build instructions?)

SecurityCRob: That's an interesting add.  I can write up an issue for consideration to address adding it

SecurityCRob: https://github.com/ossf/security-baseline/issues/472

eddie-knight: I agree with Evan's point. 

I understand mapping from build docs ->to-> user guides, because build docs could in some (probably rare) cases be categorized as a subset of user guides. It would be a reasonable, but low strength mapping.

But not the other way around.

Typical "user guides for all basic functionality" won't move you any closer to satisfying "Information on how to build all software assets"

I recommend dropping this mapping entry. But @SecurityCRob, please help me understand if I'm thinking about this the wrong way.

funnelfiasco: I am also in favor of dropping this entry.

funnelfiasco (changes requested):

baseline/OSPS-DO.yaml
SecurityCRob: This mapping is carried over from the original source from BSI.  I am not in favour of removing their work without consulting with them.  BSI BR.01 --> OSPS-DO-01

PR #464, opened 2026-01-13
Update reference-ids in OSPS-GV.yaml

SecurityCRob:
bsi mappings to gv

### DIFF
--- a/baseline/OSPS-GV.yaml
+++ b/baseline/OSPS-GV.yaml
@@ -58,7 +58,7 @@ controls:
           - reference-id: PM-30          
       - reference-id: UKSSCOP
         entries:
-          - reference-id: Claim 2.1.1 
+          - reference-id: Claim 2.1.1  
     assessment-requirements:
       - id: OSPS-GV-01.01
         text: |
@@ -116,6 +116,9 @@ controls:
           - reference-id: AC-21
           - reference-id: AU-6
           - reference-id: PL-1          
+      - reference-id: BSI-TR-03185-2
+        entries:
+          - reference-id: QA.03                 
     assessment-requirements:
       - id: OSPS-GV-02.01
         text: |
@@ -174,6 +177,11 @@ controls:
       - reference-id: UKSSCOP
         entries:
           - reference-id: Claim 2.1.1
+      - reference-id: BSI-TR-03185-2
+        entries:
+          - reference-id: GV.01
+          - reference-id: QA.04
+          - reference-id: QA.05       
     assessment-requirements:
       - id: OSPS-GV-03.01
         text: |
--- a/baseline/OSPS-GV.yaml
+++ b/baseline/OSPS-GV.yaml
@@ -58,7 +58,7 @@ controls:
           - reference-id: PM-30          
       - reference-id: UKSSCOP
         entries:
-          - reference-id: Claim 2.1.1  
+          - reference-id: Claim 2.1.1
     assessment-requirements:
       - id: OSPS-GV-01.01
         text: |


SecurityCRob:
depends on:

related to:

baseline/OSPS-GV.yaml
```diff
@@ -174,6 +177,11 @@ controls:
       - reference-id: UKSSCOP
         entries:
           - reference-id: Claim 2.1.1
+      - reference-id: BSI-TR-03185-2
+        entries:
+          - reference-id: GV.01
+          - reference-id: QA.04
```
Satarupa22-SD, line 183: I feel QA.04 is not aligned in this case. Apart from this the remaining reference mappings are consistent. CC: @eddie-knight

SecurityCRob: OSPS-GV-03 is explicitly called out by BSI when they did their mapping.  I would prefer not to adjust their work without their inclusion.

eddie-knight: @SecurityCRob I'm curious about that comment... I typically expect the document to contain outgoing mappings. So this is our claim that completing this will give some level of completion to the other control. Any claims they make are independent, right?

SecurityCRob: yes, fulfilling osps controls helps support downstream in defending their compliance to auditors.  "Compliance" is always in the eyes of the auditor, but the mappings reflect where tasks are aligned or support that defense.

baseline/OSPS-GV.yaml
```diff
@@ -58,7 +58,7 @@ controls:
           - reference-id: PM-30          
       - reference-id: UKSSCOP
         entries:
-          - reference-id: Claim 2.1.1 
+          - reference-id: Claim 2.1.1
```
evankanderson: Why add an extra space here?

baseline/OSPS-GV.yaml
evankanderson: I think "explaining the contribution process" is aligned with "procedures for testing", presuming that the project documents running the tests when contributing code.

I'd suggest there might also be a mapping to BR.01 ("how to build software assets") here, implicitly.

evankanderson (commented):
I'd argue that GV-04 ("code contributors reviewed prior to granting escalated permissions") should have a mapping to GV.02 ("the project's ... MUST be protected against unauthorized actions"), even though it's not in the BSI TR-03185-2 document.

baseline/OSPS-GV.yaml
evankanderson: See https://github.com/ossf/security-baseline/pull/460#issuecomment-3826153356

The BSI mappings are based on the 2025-02-25 version, so they don't "see" any changes or improvements we made either on 2025-10-10 or since then.  This means that we _can't_ really directly say that the BSI mappings are "authoritative", because they are pointed at something immutable in the past, rather than the next-published version, which is what these PRs go into.

baseline/OSPS-GV.yaml
eddie-knight:
```suggestion
          - reference-id: Claim 2.1.1
```

baseline/OSPS-GV.yaml
funnelfiasco: I do think we ought to be opinionated about our mappings, even when that opinion may conflict with the opinion of the organization on the other end of the mapping.

For this specific case, I feel like the link is a little tenuous, but not indefensible. I'm okay with keeping this even though it's not the choice I would make.

funnelfiasco (approved):

baseline/OSPS-GV.yaml
funnelfiasco: Given our decision in #475, we are good to proceed with this.

eddie-knight (approved):

PR #465, opened 2026-01-13
Add BSI-TR-03185-2 reference IDs to OSPS-LE.yaml

SecurityCRob:
legal things for bsi mappings

### DIFF
--- a/baseline/OSPS-LE.yaml
+++ b/baseline/OSPS-LE.yaml
@@ -96,6 +96,9 @@ controls:
       - reference-id: 800-161
         entries:
           - reference-id: PL-4                 
+      - reference-id: BSI-TR-03185-2
+        entries:
+          - reference-id: LE.01
     assessment-requirements:
       - id: OSPS-LE-02.01
         text: |
@@ -160,6 +163,9 @@ controls:
       - reference-id: 800-161
         entries:
           - reference-id: PL-4                 
+      - reference-id: BSI-TR-03185-2
+        entries:
+          - reference-id: LE.02
     assessment-requirements:
       - id: OSPS-LE-03.01
         text: |


baseline/OSPS-LE.yaml
```diff
@@ -96,6 +96,9 @@ controls:
       - reference-id: 800-161
         entries:
           - reference-id: PL-4                 
+      - reference-id: BSI-TR-03185-2
+        entries:
+          - reference-id: LE.01
     assessment-requirements:
       - id: OSPS-LE-02.01
```
Satarupa22-SD, line 103: OSPS-LE-03.02 seems more aligned in this case.  @eddie-knight 

_" [OSPS-LE-03.02](https://baseline.openssf.org/versions/2025-02-25.html#osps-le-0302): While active, the [license](https://baseline.openssf.org/versions/2025-02-25.html#license) for the [released software assets](https://baseline.openssf.org/versions/2025-02-25.html#released-software-asset) MUST be included in the released source code, or in a [LICENSE](https://baseline.openssf.org/versions/2025-02-25.html#license) file, COPYING file, or [LICENSE](https://baseline.openssf.org/versions/2025-02-25.html#license)/ directory alongside the corresponding [release](https://baseline.openssf.org/versions/2025-02-25.html#release) assets. "_

SecurityCRob: LE.01 is part of the original BSI mappings and I would encourage us not to adjust without their input.  We certainly can add LE-03 as part of our assessment for the mapping as well.

eddie-knight: Bit confused; same comment as the other PR: https://github.com/ossf/security-baseline/pull/464#discussion_r2743937579

SecurityCRob: Correct, the same comment was made for both prs.  I took whatever the BSI wrote for mappings and added items I felt were missing.  I do not desire to remove what they wrote as the basis.  If the team feels strongly about it they could reach out to the BSI and attempt to persuade them to adjust the source material that then later can be reflected in OSPS.

eddie-knight: Why are we trying to replicate their mappings as our own, instead of creating our own outgoing mappings with our own claims regarding how well our entries address theirs?

My comment on both PRs is that we should treat these mappings as our own statements.

evankanderson: The BSI mapping looks correct.

OSPS-LE-02 says that the licenses must (be stated and) meet the OSI/FSF definitions.

OSPS-LE-03 says that the licenses must be _distributed_ in standard locations (matching LE.02's requirement for a copy of the license to be distributed).


As an example, a repo which has MIT header on every file but doesn't include a copy of the MIT license would technically pass OSPS-LE-02/BSI LE.01 (though it might trip up automation), but would fail OSPS-LE-03 / BSI LE.02.

evankanderson (approved):

funnelfiasco (approved):

PR #466, opened 2026-01-13
Add BSI-TR-03185-2 references to OSPS-QA.yaml

SecurityCRob:
BSI mappings

### DIFF
--- a/baseline/OSPS-QA.yaml
+++ b/baseline/OSPS-QA.yaml
@@ -271,6 +271,9 @@ controls:
         entries:
           - reference-id: Claim 1.3.2
           - reference-id: Claim 1.3.3
+      - reference-id: BSI-TR-03185-2
+        entries:
+          - reference-id: QA.04
     assessment-requirements:
       - id: OSPS-QA-03.01
         text: |
--- a/baseline/OSPS-QA.yaml
+++ b/baseline/OSPS-QA.yaml
@@ -72,6 +72,9 @@ controls:
       - reference-id: UKSSCOP
         entries:
           - reference-id: Claim 2.2.3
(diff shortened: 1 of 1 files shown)


baseline/OSPS-QA.yaml
```diff
@@ -468,6 +477,10 @@ controls:
           - reference-id: SA-11
           - reference-id: SA-15
           - reference-id: SR-3                
+      - reference-id: BSI-TR-03185-2
+        entries:
+          - reference-id: QA.04
+          - reference-id: QA.05
```
evankanderson, line 485: I'd quibble about the mapping from BSI QA.05 (memory safety), but I think this mapping is fine.  (OSPS seems less concerned with memory safety than some other efforts, which seems fine to me -- no need to duplicate efforts.)

evankanderson (commented):
It seems like OSPS-QA-03 ("automated status checks MUST pass ...") could map to BSI QA.04 ("procedures for testing...").

baseline/OSPS-QA.yaml
funnelfiasco: Yeah BSI QA.05 is more specific than we are here, but I think that's okay.

funnelfiasco (dismissed):

evankanderson:
> It seems like OSPS-QA-03 ("automated status checks MUST pass ...") could map to BSI QA.04 ("procedures for testing...").

Since the QA-03 mappings aren't in the PR, I can't make a suggestion to add them, but I'd argue that having automated status checks provides at least one procedure for testing.

funnelfiasco (approved):

eddie-knight (approved):

PR #479, opened 2026-02-19
Release 2026.02.19

funnelfiasco:


docs/release_notes.md
```diff
@@ -39,4 +70,4 @@
 
 ## 2025-02-25
 
-Initial release of the Open Source Project Security Baseline.
\ No newline at end of file
+Initial release of the Open Source Project Security Baseline.
```
SecurityCRob, line 73: Is this still the "initial" release?

SecurityCRob (approved):
Excellent work team.  I approve

docs/release_notes.md
funnelfiasco: We could go back and add more details, but since 2025-02-25 is now several versions ago (and the diff is "literally all of the baseline at that point"), it doesn't seem worthwhile.

funnelfiasco: Oh, I think I see what's going on here. It seems vim dropped the newline, so this showed up in the diff when it really shouldn't have. But that's probably fine. 😅

Issue #489, opened 2026-03-23
Allow LICENSE/ and LICENSES/


david-a-wheeler:
osps_le_03_02 says, "While active, the license for the released software assets MUST be included in the released source code, or in a LICENSE file, COPYING file, or LICENSE/ directory alongside the corresponding release assets."

However, it fails to include a LICENSES/ directory. For example, this repo uses "LICENSES/": https://github.com/oss-review-toolkit/ort This is also an unnecessary incompatibility with the best practices "Metal" criteria, whose `license_location` requires that "The project MUST post the license(s) of its results in a standard location in their source repository (e.g., LICENSE or COPYING file, or LICENSES/ directory)."

I propose adding `LICENSES/` like this:

> "While active, the license for the released software assets MUST be included in the released source code, or in a LICENSE file, COPYING file, LICENSES/ directory, or LICENSE/ directory alongside the corresponding release assets."

funnelfiasco:
I'm fine with calling this a bug in the requirement. My question is if we want to continue to go down the road of adding explicit alternatives or just say something like "or similar, clearly-identifiable location" to account for, e.g. `LICENSE.md`. 

I'll note that in the case of the ort example, that project _does_ have a `LICENSE` file which contains the text of ASL-2.0. The `LICENSES/` directory appears to be the license of some subcomponents within the code. This makes it an example of a problem Fedora had, which is: "do we display every license in a project or the 'effective' license of the project?" Arguably, the _effective_ license of the project is ASL-2.0 because following that meets all of the requirements of the other licenses and so the presence of the `LICENSE` file is sufficient for Baseline conformance anyway. This is more of a problem for the tooling implementors to solve than a question for the Baseline itself, but it's worth calling out.

eddie-knight:
A previous suggestion was to use the phrase "well known location" and then shift the list to reccos

david-a-wheeler:
In this *specific* situation I think being specific is important. Many tools *only* look in specific locations, it's basically an API, so constraining them is a good plan.

But in most cases I'd agree with you. Unless we need to constrain something as an API, flexibility is good.

david-a-wheeler:
This was resolved in #498 - thank you everyone!

PR #497, opened 2026-04-16
Add LICENSES/ per issue #489

david-a-wheeler:
Many projects use LICENSES/ directory not a LICENSE/ directory, per <https://reuse.software/faq/>. While we *could* allow arbitrary directories, having a *short* list of places to look makes tooling and analysis much easier. The naming also makes more sense, because the usual *reason* to have a directory is because there is more than 1 (plural) licenses involved.

So let's allow LICENSES/ but specifically specify it.


david-a-wheeler:
Weird, there's a conflict. I'll try again.

PR #504, opened 2026-04-28
feat: add --crosswalk-output flag for reverse framework crosswalk

### LINKED ISSUE #379: Create reverse crosswalks for other frameworks
Right now, each Baseline control lists corresponding requirements in other frameworks. It would be nice if we did this in reverse: for each of the mapped frameworks, list corresponding Baseline controls.

liketosweep:
## What the Issue Was

The `baseline-compiler` only supports forward-facing generation - exporting
controls and their associated framework mappings. Auditors and security engineers
working in reverse (starting from a NIST or CIS sub-control and identifying which
OSPS controls satisfy it) had no automated path. Manually tracing these
relationships through YAML source files is inefficient and error-prone.

## Why It Was There

The generation logic was previously coupled with standard template rendering, with
no inversion step. The data structure was `Control -> []Requirement`, with no
mechanism to pivot to a `Framework -> Requirement -> []Control_ID` hierarchy
needed for audit crosswalk artifacts. There was also no CLI entry point to trigger
such a generation path.

## Solution

Introduced a `--crosswalk-output` string flag to the `compile` command. When a
path is provided, `RunE` triggers a dedicated crosswalk generator rather than the
standard template renderer. The generator inverts the mapping data via
`buildReverseCrosswalk` and writes the result using `ExportReverseCrosswalk` with
an `io.Writer` for flexibility. Lexicographical sorting is applied to Framework
names, Requirement IDs, and Control IDs to ensure byte-for-byte stable output
across runs and environments.

## Changes

- **`cmd/internal/cmd/compile.go`** - Introduced `--crosswalk-output` string flag;
  added logic in `RunE` to trigger the crosswalk generator when a path is provided;
  added a success indicator (✅) to terminal output consistent with existing UX.

- **`pkg/baseline/generator_crosswalk.go`** - Created `buildReverseCrosswalk` to
  perform mapping inversion; implemented `ExportReverseCrosswalk` via `io.Writer`;
  applied deterministic lexicographical sorting across all output dimensions.

- **`pkg/baseline/generator_crosswalk_test.go`** - Added unit tests using the
  project's internal `types` to mock a baseline catalog; verified correct grouping
  of requirements under frameworks and byte-for-byte output stability across runs.

## Result

The `baseline-compiler` can now generate a reverse crosswalk artifact as a
first-class CLI operation. Auditors can map from any regulatory requirement
(NIST, CIS) back to the OSPS controls that satisfy it  - automatically, and as
part of any CI/CD pipeline or local dev workflow - without manually tracing YAML
source files.

## Conducted Tests 

Testing Note: Verified generation logic via 100% unit test coverage. During manual CLI testing on Windows (MinGW64), I discovered that the existing loader.go currently fails during local URI fetching (unsupported URI scheme / invalid port errors) because fetcher.URI{} struggles to parse native Windows file paths. I have kept this PR strictly scoped to the crosswalk generation logic, but I would love to tackle that cross-platform URI bug in a separate issue!

image

---

### Example Output
Running `./baseline-compiler compile --crosswalk-output reverse.md` generates the following artifact structure:

| Framework | Requirement | OSPS Controls |
| :--- | :--- | :--- |
| CIS Controls v8 | 1.1 | OSPS-GV-01 |
| NIST SP 800-53 | AC-2 | OSPS-AC-01, OSPS-GV-01 |
| NIST SP 800-53 | AC-3 | OSPS-AC-01 |

Resolves #379

### DIFF
--- a/cmd/internal/cmd/compile.go
+++ b/cmd/internal/cmd/compile.go
@@ -152,7 +152,11 @@ func addCompile(parentCmd *cobra.Command) {
 				if err != nil {
 					return fmt.Errorf("creating crosswalk output file: %w", err)
 				}
-				defer f.Close()
+				defer func() {
+					if cerr := f.Close(); cerr != nil {
+						fmt.Printf("error closing file: %v\n", cerr)
+					}
+				}()
 
 				// Use the existing 'gen' and 'bline' variables
 				if err := gen.ExportReverseCrosswalk(bline, f); err != nil {
--- a/cmd/internal/cmd/compile.go
+++ b/cmd/internal/cmd/compile.go
@@ -17,6 +17,7 @@ type compileOptions struct {
 	outPath               string
 	checklistOutPath      string
 	baselinePath          string
+	crosswalkOutputPath   string
 	checklistTemplatePath string
 	templatePath          string
 	validate              bool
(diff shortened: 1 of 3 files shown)


funnelfiasco (approved):
Looks good. Thanks for taking this on.

funnelfiasco:
@kusari-inspector rerun

SecurityCRob (approved):

PR #535, opened 2026-08-25
ci: publish catalog and mapping documents to grc.store on release

eddie-knight:
Adds `.github/workflows/publish.yaml`, which publishes the OSPS Baseline to [grc.store](https://grc.store) as versioned, signed OCI artifacts.

On a published GitHub Release it compiles the catalog with `baseline-compiler gemara`, CUE-vets the result and all 14 mapping documents, stamps each artifact's `metadata.version` from the release tag (staged copies — committed YAML is untouched), and publishes each one with `grcli`. Auth is OIDC trusted publishing: `id-token: write` is the whole setup, no secrets. Signing is keyless and happens in-process inside `grcli` via sigstore-go — there is no `cosign` install step, since `cosign` is only needed for the `--cosign-key` path we don't use.

Manual dispatch is also supported and defaults to a dry run, which builds the OCI layouts locally and uploads them as a build artifact.

### Also in this PR

- `.github/actions/compile-and-vet` — the build/compile/CUE-vet steps were duplicated verbatim between this workflow and `cue-validate.yaml`. Two copies of the CUE gate drift silently, and that gate is what stands between a bad mapping document and a signed publish, so both now call one composite action. `cue-validate.yaml` shrinks from 49 lines to 26.
- `.github/ISSUE_TEMPLATE/release.yaml` — the release checklist ended at "git tag the commit". The trigger here is `release: published`, so a bare tag would never have fired this workflow. Added the missing step.

### Before it can publish for real

Neither is a code change, so merging ahead of them is safe — an unregistered publisher just gets a 403:

1. Register `ossf/security-baseline` as an OIDC trusted publisher for the `openssf` namespace on hub.grc.store. All 15 artifacts carry `metadata.author.id: openssf` with distinct `metadata.id`, so one namespace binding covers `openssf/osps-baseline` and the 14 `openssf/osps-baseline-to-*` repositories.
2. Create an actual GitHub Release — see the checklist change above.

Suggested first run after merge: a manual dispatch with `dry_run` left at its default, to confirm the compile → CUE gate → layout pipeline end to end.

### Notes for reviewers

- **`grcli` is pinned by digest, not tag.** GHCR tags are mutable, and this is the one binary that executes holding an OIDC publishing credential.
- **The pull needs no credentials.** `ghcr.io/gemaraproj/grcli` is public. A raw unauthenticated request to ghcr.io returns 401, but that is the `WWW-Authenticate` challenge — ghcr.io issues anonymous pull tokens and `oras` follows the challenge automatically. Verified with a credential-free `oras pull` of the pinned digest.
- **One `grcli publish` per artifact, deliberately.** Positional args to `grcli publish` describe a *single* artifact, so batching all 15 into one call would produce one merged bundle rather than 15.
- **Each artifact gets its own `--output`.** `grcli`'s dry run writes an OCI layout tagged with `metadata.version` alone, with no repository component. Since every artifact in a release shares a version, a single shared output directory would leave only the last one addressable by tag.
- **One `cue vet` per mapping document, deliberately.** Vetting them together would unify all 14 into a single CUE instance instead of checking each on its own.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -8,11 +8,11 @@ name: Publish to grc.store
 # grcli. Triggered automatically when a GitHub Release is published; also
 # runnable manually for testing (dry_run defaults to true).
 #
-# Requirements:
-#   - Repository secret GRCLI_TOKEN: bearer token for the hub sync call.
-#     Obtain by running `grcli login` locally and copying the access token from
-#     ${XDG_DATA_HOME:-~/.local/share}/grcli/credentials.json, or by issuing a
-#     service-account token in the grc.store hub UI.
+# Auth:
+#   - grcli authenticates to grc.store via OIDC trusted publishing. The
+#     id-token: write permission below lets the runner mint an OIDC token
+#     that grcli exchanges for a hub credential. No static secrets needed —
+#     register this repo as a trusted publisher in the grc.store hub UI.
 #
 # Notes:
 #   - grcli is distributed as an OCI artifact (not a runnable container),
@@ -37,6 +37,7 @@ on:
 permissions:
   contents: read
   packages: read
+  id-token: write
 
 jobs:
   publish:
@@ -81,7 +82,6 @@ jobs:
 
       - name: Publish catalog and mapping documents to grc.store
         env:
-          GRCLI_TOKEN: ${{ secrets.GRCLI_TOKEN }}
           DRY_RUN: ${{ github.event.inputs.dry_run }}
           EVENT_NAME: ${{ github.event_name }}
           RELEASE_TAG: ${{ github.event.release.tag_name }}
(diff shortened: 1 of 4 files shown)


funnelfiasco (approved):

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,87 @@
...
+      # workflow that checks a PR and the one that publishes.
+      - name: Compile catalog and vet it against the CUE schema
+        uses: ./.github/actions/compile-and-vet
+
+      - name: Setup oras
+        uses: oras-project/setup-oras@8d34698a59f5ffe24821f0b48ab62a3de8b64b20 # v1.2.3
+
+      - name: Install grcli
+        run: |
+          oras pull ghcr.io/gemaraproj/grcli@sha256:9e668b4a5a6bc93c48190f2495911e20276bbc3dff9d9e0b0de35a859b69ca5f --platform linux/amd64
```
evankanderson, line 63: Can/will dependabot update these image references?

evankanderson (approved):

.github/workflows/publish.yaml
eddie-knight: I don't believe so, but the articles I'm looking at could be out of date

PR #540, opened 2026-08-26
fix: lexicon framework terms

eddie-knight:
This removes duplicate metadata/lexicon entries and removes references that are not in use. 

Also adjusts the template to accommodate:

- :120 switches the link source from lexicon lookup to a direct anchor — this is what replaces the deleted entries.
- :136 creates the anchor that line points at. Nothing could link into the External Frameworks table before; every row was unaddressable. Without this line, :120 produces dead links instead of working ones.

Edit: Claude recommended some CI changes to avoid drift in the future. See below for that info.

🤖 

This removes duplicate metadata/lexicon entries and references that are not in use, and hardens the checks that keep them from coming back.

## What the template change actually fixes

Framework relations rendered through `addLinks`, which produced reference-style links into the **lexicon glossary** rather than the External Frameworks table. Only 9 of the 14 frameworks linked at all, and two of those linked wrongly:

- `**BSI-TR-03185-2**`, `**PSSCRM**`, `**Scorecard**`, `**UKSSCOP**` rendered as bare bold text, having no lexicon entry.
- `**ISO-[18974][OpenChain]**` was mangled. The OpenChain entry carried the synonym `18974`, and `addLinks` substring-matched it *inside* the identifier `ISO-18974`, splitting the ID and emitting a link in the middle of it.
- `**[SAMM][Proactive Software Supply Chain Risk Management Framework]**` pointed at the **wrong framework**. That title was the duplicated lexicon term, so SAMM's link resolved to P-SSCRM's glossary anchor — and the duplicate caused the same link definition to be emitted twice in one document.

So the duplicate lexicon entry was not cosmetic: it produced a wrong link destination in published output.

The template changes:

- `:120` switches the link source from lexicon lookup to a direct anchor — this is what replaces the deleted entries.
- `:136` creates the anchor that line points at. Nothing could link into the External Frameworks table before; every row was unaddressable. Without this line, `:120` produces dead links instead of working ones.

All 14 frameworks now render uniformly as `[ID](#fw-ID)` into the table, with no reference-style or mangled links remaining.

## Validation

Routing framework names through the lexicon was the wrong mechanism — a short synonym could corrupt a longer identifier. Direct anchors remove that class of bug, and the new checks keep the data behind them honest:

- A mapping document must target a mapping-reference declared in the catalog metadata. The empty string is **not** exempt: it is the one value that fails invisibly, since both `validate` and `cue vet` pass while the renderer drops the document from the relations and the crosswalk in silence.
- The reverse: a mapping-reference declared in metadata with no mapping document behind it renders a row nothing can link to.
- Lexicon names are compared case-folded and across synonyms, since `addLinks` resolves a name to the first entry declaring it and `asLink` folds terms into anchors. This required dropping two more synonyms: `Vulnerability Reporting` claimed `Coordinated Vulnerability Disclosure`, which is its own entry with a different definition, and the SLSA entry repeated its own term.
- The reference ID shape is constrained in CUE, on both the catalog metadata and each mapping document's target reference, so an ID containing a space cannot render a broken anchor.

Both new checks are covered by tests. Rendered output is byte-identical to before the validation changes, and anchors remain 14/14.

### DIFF
--- a/.project-words.txt
+++ b/.project-words.txt
@@ -1,6 +1,5 @@
 CCCS
 CISA
-CSAG
 crosswalked
 devel
 DSIT
@@ -42,6 +41,5 @@ Subprojects
 triaging
 unreviewable
 UKSSCOP
-USCTM
 Updegrove
 webfonts
--- a/baseline/lexicon.yaml
+++ b/baseline/lexicon.yaml
@@ -36,14 +36,6 @@
     An automated test suite must return an overall "pass" or "fail" result,
     and is often implemented using a test framework.
     Common ways to invoke automated tests include `make check`, `make test`, `npm test`, and `cargo test` manually or as part of a Continuous Integration workflow.
-- term: Best Practices Badge
-  definition: |
-     The OpenSSF Best Practices Badge Identifies FLOSS best practices & implements a badging system for those practices.
-  synonyms: 
-    - BPB
-    - OpenSSF Best Practices Badge
-  references: 
-    -  https://www.bestpractices.dev/en
 - term: Build and Release Pipeline
   definition: |
     A series of automated processes that compile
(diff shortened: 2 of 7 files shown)


funnelfiasco (dismissed):

funnelfiasco (approved):

SecurityCRob (approved):
+1

evankanderson (approved):
It looks like this changes the format of anchor references, but that should only take effect for *new* baseline versions because of the way we check in the compiler outputs for existing versions.

PR #544, opened 2026-08-28
Remove "While active" qualifier from baseline

### LINKED ISSUE #398: "while active" felt confusing
Reading the baseline entries, the "while active" items felt somewhat arbitrary (and unclear as to its meaning - EOL? Developers haven't done anything for 2 weeks?).

For example, an inactive project would fail baseline level 1 if it has an 'http:' url, but it is fine if the project's source is no longer available.

Note also that items like 'OSPS-QA-01.02' appear to imply 'while active' as they wouldn't make sense if OSPS-QA-01.01 is passed only because the project isn't active.

Satarupa22-SD:
As per the [style guide](docs/style-guide.md) we should remove _while Active_ qualifier across the baselines.
CC: @eddie-knight @funnelfiasco

### DIFF
--- a/baseline/OSPS-GV.yaml
+++ b/baseline/OSPS-GV.yaml
@@ -20,7 +20,7 @@ controls:
       - id: OSPS-GV-01.01
         text: |
          The project documentation MUST include a list of project members with 
-         access to sensitive resources
+         access to sensitive resources.
         applicability:
           - maturity-2
           - maturity-3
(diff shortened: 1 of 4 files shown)


baseline/OSPS-GV.yaml
```diff
@@ -19,8 +19,8 @@ controls:
     assessment-requirements:
       - id: OSPS-GV-01.01
         text: |
-          While active, the project documentation MUST include a list of
-          project members with access to sensitive resources.
+         The project documentation MUST include a list of project members with 
+         access to sensitive resources
```
eddie-knight:
```suggestion
         access to sensitive resources.
```

eddie-knight (dismissed):

funnelfiasco (approved):

eddie-knight (approved):

Issue #546, opened 2026-08-28
Audit and split requirements combining multiple obligations


Satarupa22-SD:
As per the [style guide](docs/style-guide.md):

> **A Single Requirement corresponds to a single Test**
> Every requirement MUST describe exactly one testable behavior. Each requirement should map directly to a single implementation test. If a requirement has multiple obligations that can pass/fail independently, split them into separate requirements. An easy way to enforce this is to avoid using "and" in the requirement text.

Several requirements currently join two or more independently-testable obligations with "and" (e.g., "evaluated ... **and** blocked", "document **when and how**"). A project could satisfy one clause while failing the other, so these need to be reviewed and, where applicable, split so each requirement maps to a single test.

Precedent: [#443](https://github.com/ossf/security-baseline/pull/443) split `BR-01` into three requirements.

### Requirements to review

- [x] [OSPS-VM-06.02](https://baseline.openssf.org/versions/2026-02-19#osps-vm-0602) 
- [x] [OSPS-VM-05.03](https://baseline.openssf.org/versions/2026-02-19#osps-vm-0503)
- [x] [OSPS-VM-05.01](https://baseline.openssf.org/versions/2026-02-19#osps-vm-0501) 
- [x] [OSPS-VM-01.01](https://baseline.openssf.org/versions/2026-02-19#osps-vm-0101)
- [x] [OSPS-QA-06.02](https://baseline.openssf.org/versions/2026-02-19#osps-qa-0602) 

CC: @eddie-knight @funnelfiasco

PR #548, opened 2026-08-28
feat: update OSPS-VM-06

Satarupa22-SD:
refer #546

### DIFF
--- a/baseline/OSPS-VM.yaml
+++ b/baseline/OSPS-VM.yaml
@@ -178,14 +178,31 @@ controls:
           these findings.
       - id: OSPS-VM-06.02
         text: |
-          All changes to the project's codebase MUST be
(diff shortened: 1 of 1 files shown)


Satarupa22-SD:
@funnelfiasco @eddie-knight, please review this. Thank you!

evankanderson (changes requested):
I realize the existing statement is long, but it seems coherent as a single policy:

> The project must have an automated enforcement policy to prevent the introduction of security weaknesses into the codebase.

A few details:

1. The existence of the enforcement automation serves as _de facto_ documentation, so calling it a "documented policy" may be duplicative if enforcement is required.
2. This statement does not explicitly declare that the enforcement policy can permit "ignore" or "suppress" mechanisms, but it also does not restrict them, so we can use the recommendation to mention that possibility.

Issue #543, opened 2026-08-27
[Release]: 2026.08.28


eddie-knight:
### Version

2026.08.28

### Release checklist

- [x] Ensure all pull requests targeted for this release are merged
- [x] Generate a static version of the baseline: `cd cmd && go run . compile --output ../docs/versions/2026-08-28.md --checklist-output ../docs/versions/2026-08-28-checklist.md --crosswalk-output ../docs/versions/2026-08-28-crosswalk.md --artifact-version v2026.08.28`
- [x] In the `docs/versions/<PREVIOUS_VERSION>.md` file, remove the nav-title header block
- [x] In the `docs/versions/<VERSION>.md` file, change the `Version: devel` to have the version number of the new version
- [x] In the `docs/versions/<VERSION>.md` file, remove the "not for production use" warning
- [x] In the `docs/versions/<VERSION>.md` file, add the nav-title
- [x] In the `docs/_config.yml` file, update the listing in the `nav_pages` group to point to the new release
- [x] In the `docs/versions/<VERSION>-checklist.md` and `docs/versions/<VERSION>-crosswalk.md` files, change the headers to have the version number of the new version
- [x] Move the link to the "current" version in `docs/index.md` to the "previous versions" list
- [x] Add the link to the new version to the "current version" in `docs/index.md`
- [x] Make sure the `docs/versions/release_notes.md` file contains an accurate list of changes and is not commented out
- [x] Open a pull request with the above changes (#542)
- [x] After the pull request is merged, git tag the commit
- [x] Publish a GitHub Release from that tag — this is what triggers the grc.store publish workflow; a bare git tag will not fire it
- [ ] After the pull request is merged, send an email to the [mailing list](https://lists.openssf.org/g/openssf-sig-security-baseline)
- [ ] After the pull request is merged, notify OpenSSF's marketing team that we have a new release so that they can share it with the world

PR #549, opened 2026-08-31
fix: regenerate v2026.08.28 pages with mid-cut baseline changes

eddie-knight:
A few PRs came across the finish line after the website update in #543 

Because the website content update is not automatic, the release tag and artifacts for v2026.08.28 are out of sync with the website. This PR adds those changes.

#551 is drafted to track the automation gap.

### DIFF
--- a/docs/release_notes.md
+++ b/docs/release_notes.md
@@ -44,6 +44,8 @@
 ### Modified controls
 
 * [OSPS-LE-03.01](versions/2026-08-28#osps-le-0301) now also accepts a `LICENSES/` directory as a license location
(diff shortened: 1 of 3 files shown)


funnelfiasco (approved):

puerco (approved):

david-a-wheeler (approved):
Looks good. Saying criteria only apply "when active" would mean an inactive project wouldn't need to meet any requirement, which is dangerous.

SecurityCRob (commented):
Lgtm, ty

PR #550, opened 2026-08-31
fix(ci): stamp baseline target version and draft flag at publish time

eddie-knight:
Published artifacts for v2026.08.28 shipped with their osps-baseline mapping-reference still reading 'draft' and metadata.draft still true, because only metadata.version was stamped at publish time. The committed YAML is correct as-is (in-repo it tracks devel), so extend the existing publish-time stamp instead of editing the documents each release:

- pin each mapping document's osps-baseline reference to RELEASE_VERSION
- set metadata.draft from ARTIFACT_DRAFT: false for explicitly versioned publishes (release tag or dispatch version input), true for the dev-SHA fallback; only applied where a draft flag already exists

All 15 stamped artifacts verified against the CUE schema locally.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,7 +6,6 @@ docs/_site
 docs/versions/devel-checklist.md
 docs/versions/devel-crosswalk.md
 cmd/baseline.gemara.*
-cmd/.*
 
 # go build artifacts
 cmd/security-baseline
--- a/.gitignore
+++ b/.gitignore
@@ -2,9 +2,11 @@ docs/versions/devel.md
 .DS_Store
 docs/_site
 
-# generated output from go run ./... compile
+# generated output
 docs/versions/devel-checklist.md
 docs/versions/devel-crosswalk.md
+cmd/baseline.gemara.*
+cmd/.*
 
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -2,9 +2,11 @@ docs/versions/devel.md
 .DS_Store
 docs/_site
 
-# generated output from go run ./... compile
+# generated output
 docs/versions/devel-checklist.md
 docs/versions/devel-crosswalk.md
+cmd/baseline.gemara.*
+cmd/.*
```
funnelfiasco: I'm a little hesitant to exclude all dotfiles from the directory because it could lead to some unexpected behavior if we ever need a dotfile in this directory. Can we be more explicit here?

eddie-knight: Sure, it's safe to remove that line outright. I just saw that my claude was shoving swap files in there while working and figured I'd add an outright exclusion. Not necessary.

```suggestion
```

funnelfiasco (approved):

david-a-wheeler (approved):

PR #553, opened 2026-09-03
Update VM-05 to split dependency remediation enforcement into 4 areas

Satarupa22-SD:
refer #546

### DIFF
--- a/baseline/OSPS-VM.yaml
+++ b/baseline/OSPS-VM.yaml
@@ -143,7 +143,7 @@ controls:
           that verify compliance with that policy prior to release.
       - id: OSPS-VM-05.03
         text: |
-          Retired in <PR_URL>
+          Retired in https://github.com/ossf/security-baseline/pull/553
         applicability:
           - maturity-3
         state: Retired
--- a/baseline/OSPS-VM.yaml
+++ b/baseline/OSPS-VM.yaml
@@ -147,7 +147,7 @@ controls:
         applicability:
           - maturity-3
         state: Retired
-      - id: OSPS-VM-05.04
+      - id: OSPS-VM-05.06
         text: |
           All changes to the project's codebase MUST be
           automatically evaluated against a documented policy for malicious
@@ -159,7 +159,7 @@ controls:
           runs a Software Composition Analysis tool to detect malicious
           dependencies present in the codebase. Require that the status check
           passes before any new changes can be merged.
-      - id: OSPS-VM-05.05
+      - id: OSPS-VM-05.07
         text: |
           All changes to the project's codebase MUST be
           automatically evaluated against a documented policy for known
@@ -171,7 +171,7 @@ controls:
           runs a Software Composition Analysis tool to detect known
           vulnerabilities in dependencies. Require that the status check
           passes before any new changes can be merged.
-      - id: OSPS-VM-05.06
+      - id: OSPS-VM-05.08
         text: |
           Changes to the codebase MUST be blocked when they violate the
           documented policy for malicious dependencies or known
@@ -182,7 +182,7 @@ controls:
           Configure status checks to block any new changes from being merged
           when the documented policy for malicious dependencies or known
           vulnerabilities is violated.
-      - id: OSPS-VM-05.07
+      - id: OSPS-VM-05.09
         text: |
           A project MUST allow documented dependency policy violations to be
           suppressed when they are declared non-exploitable.
--- a/baseline/OSPS-VM.yaml
+++ b/baseline/OSPS-VM.yaml
@@ -143,18 +143,57 @@ controls:
           that verify compliance with that policy prior to release.
       - id: OSPS-VM-05.03
         text: |
+          Retired in <PR_URL>
+        applicability:
+          - maturity-3
+        state: Retired
+      - id: OSPS-VM-05.04
+        text: |
           All changes to the project's codebase MUST be
           automatically evaluated against a documented policy for malicious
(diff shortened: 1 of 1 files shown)


baseline/OSPS-VM.yaml
```diff
@@ -142,19 +142,58 @@ controls:
...
           - maturity-3
         recommendation: |
           Create a status check in the project's version control system that
-          runs a Software Composition Analysis tool on all changes
-          to the codebase. Require that the status check passes before changes
-          can be merged.
+          runs a Software Composition Analysis tool to detect malicious
+          dependencies present in the codebase. Require that the status check
+          passes before any new changes can be merged.
+      - id: OSPS-VM-05.07
```
funnelfiasco, line 162: Shouldn't this be VM-05.04?

baseline/OSPS-VM.yaml
```diff
@@ -142,19 +142,58 @@ controls:
...
+          automatically evaluated against a documented policy for known
+          vulnerabilities in dependencies.
+        applicability:
+          - maturity-3
+        recommendation: |
+          Create a status check in the project's version control system that
+          runs a Software Composition Analysis tool to detect known
+          vulnerabilities in dependencies. Require that the status check
+          passes before any new changes can be merged.
+      - id: OSPS-VM-05.08
```
funnelfiasco, line 174: Shouldn't this be VM-05.05?

funnelfiasco (changes requested):
Seems like we have a gap in our numbering?

funnelfiasco:
Beyond my numbering comments, I wonder if we really want two controls here or if we we just go with "Changes to the codebase MUST be blocked when they violate the documented policy for malicious dependencies or known vulnerabilities." and let that imply the scanning in the previous control.

This keeps the count lower. The failure mode, of course, is that someone says "well if we don't scan the changes, we don't have to worry about it." I'm not sure we want to spend a lot of time preventing malicious compliance, though. This is a broader question for the team to work out, not an issue with this PR specifically.

baseline/OSPS-VM.yaml
```diff
@@ -142,19 +142,58 @@ controls:
...
+        applicability:
+          - maturity-3
+        recommendation: |
+          Configure status checks to block any new changes from being merged
+          when the documented policy for malicious dependencies or known
+          vulnerabilities is violated.
+      - id: OSPS-VM-05.09
+        text: |
+          A project MUST allow documented dependency policy violations to be
+          suppressed when they are declared non-exploitable.
```
evankanderson, line 188: I'm not sure that this is a necessary control.  I don't think we want to _forbid_ this behavior (I've found it useful in a variety of cases), but this seems more like a "taste" thing than a requirement.

baseline/OSPS-VM.yaml
```diff
@@ -142,19 +142,58 @@ controls:
...
+        recommendation: |
+          Create a status check in the project's version control system that
+          runs a Software Composition Analysis tool to detect known
+          vulnerabilities in dependencies. Require that the status check
+          passes before any new changes can be merged.
+      - id: OSPS-VM-05.08
+        text: |
+          Changes to the codebase MUST be blocked when they violate the
+          documented policy for malicious dependencies or known
+          vulnerabilities.
```
evankanderson, line 178: I'm not sure how I feel about breaking the previous (admittedly wordy) control up into three, where the three are:

1. Block malicious dependencies
2. Block vulnerable dependencies
3. The blocking must be automatic and enforced

I can see separating 1 and 2 (though I could also see arguing that malicious dependencies are de-facto "vulnerable"), but splitting this into 3 or 4 seems too much.

evankanderson (commented):
I feel like "a documented policy" leaves plenty of space for "a policy which permits certain types of vulnerabilities, including non-exploitable ones" and possibly other policy which even permit certain types of vulnerabilities that are low-severity (e.g. ReDoS).

I'm okay with splitting this into two for simplification purposes; there are a mix of "ands" and "or"s in the sentence that makes it somewhat harder to read.

baseline/OSPS-VM.yaml
puerco: This is extracted from the previous wording which was intended to still allow compliance when you VEX a non-exploitable vulnerability:

> except when declared and suppressed as non-exploitable.

Now it reads as _"you can cheat on your policy when you document the cheat"_:

> allow documented dependency policy violations

baseline/OSPS-VM.yaml
Satarupa22-SD: Both of them are already in use here: https://github.com/ossf/security-baseline/pull/554/changes. I initially labeled them VM-05.04 and VM-05.05, but later updated them because OSPS-VM-05.01 comes first in the order, although the numbering is already somewhat disrupted.

baseline/OSPS-VM.yaml
Satarupa22-SD: Thank you for reviewing this. I agree with your comments and will update the wording accordingly.

baseline/OSPS-VM.yaml
eddie-knight: I think this remains true to what it's always been. Roughly...

1. Evaluate using a policy about code that introduces risk
2. Evaluate using a policy about code that introduces an attack
3. Block policy violations

evankanderson: It feels to me like "Block policy violations" should just be part of the control, unless we're planning to put the "define a policy" and "enforce a policy" at different baseline conformance levels.

PR #555, opened 2026-09-03
Split OSPS-VM-01.01 into two requirements

Satarupa22-SD:
refer #546

### DIFF
--- a/baseline/OSPS-VM.yaml
+++ b/baseline/OSPS-VM.yaml
@@ -20,9 +20,15 @@ controls:
     assessment-requirements:
       - id: OSPS-VM-01.01
         text: |
-          The project documentation MUST
-          include a policy for coordinated vulnerability disclosure (CVD), with a clear
-          timeframe for response.
+          Retired in https://github.com/ossf/security-baseline/pull/555
+        applicability:
+          - maturity-2
+          - maturity-3
+        state: Retired
+      - id: OSPS-VM-01.02
+        text: |
+          The project documentation MUST include a coordinated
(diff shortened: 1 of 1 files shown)


funnelfiasco (commented):
I don't think this is a case of needing to retire a control and introduce a new one. Ultimately, we're still asking for the same result. The timeframe for response is one required element of the CVD policy

baseline/OSPS-VM.yaml
```diff
@@ -20,9 +20,15 @@ controls:
...
-          timeframe for response.
+          Retired in https://github.com/ossf/security-baseline/pull/555
+        applicability:
+          - maturity-2
+          - maturity-3
+        state: Retired
+      - id: OSPS-VM-01.02
+        text: |
+          The project documentation MUST include a coordinated
+          vulnerability disclosure (CVD) policy.
```
eddie-knight, line 31: I can get behind Ben's idea here... We are loosening this slightly, but not changing the meaning. Adding strictness or changing meaning would demand a retirement.

```suggestion
          The project documentation MUST include a coordinated
          vulnerability disclosure (CVD) policy.
```

Satarupa22-SD: In that case, should i just update the existing control instead of splitting this?

evankanderson (changes requested):
I'm with Ben here -- I think this is one check, rather than "is there a policy" and "does the policy contain certain content".  More granularity and more checks is not necessarily better, even if we have a few checks which are too big.

PR #559, opened 2026-09-07
Update OSPS-BR-01.03

Satarupa22-SD:
Removes from the requirement description.
CC: @funnelfiasco @eddie-knight


eddie-knight (approved):
good call

evankanderson:
This seems like a good example of where the "and" rule in the styleguide may be going too far.  The CI running untrusted code must be prevented from accessing either type of secure content:
1. credentials
2. assets such as release artifacts

Does this technically mean two separate checks, one for each type of content?  Possibly, but that seems inconvenient, so I'm inclined to keep the scope of the rule.

